PR #18, opened 2019-12-04
Fix query parameter requirements

pakrym:
.net does stricter escaping then python

### DIFF
--- a/mappings/vanilla/queries_datetime_2012-01-01t010101z-a35a5a1b-b3a4-415e-9e60-f764ef114fcb.json
+++ b/mappings/vanilla/queries_datetime_2012-01-01t010101z-a35a5a1b-b3a4-415e-9e60-f764ef114fcb.json
@@ -2,7 +2,12 @@
   "id" : "a35a5a1b-b3a4-415e-9e60-f764ef114fcb",
   "name" : "queries_datetime_2012-01-01t010101z",
   "request" : {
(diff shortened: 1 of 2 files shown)


mappings/vanilla/queries_datetime_2012-01-01t010101z-a35a5a1b-b3a4-415e-9e60-f764ef114fcb.json
```diff
@@ -2,7 +2,12 @@
   "id" : "a35a5a1b-b3a4-415e-9e60-f764ef114fcb",
   "name" : "queries_datetime_2012-01-01t010101z",
   "request" : {
-    "url" : "/queries/datetime/2012-01-01T01%3A01%3A01Z?dateTimeQuery=2012-01-01T01%3A01%3A01.000Z",
+    "urlPath" : "/queries/datetime/2012-01-01T01%3A01%3A01Z",
+    "queryParameters" : {
+      "dateTimeQuery" : {
+        "equalTo" : "2012-01-01T01:01:01Z"
```
lmazuel: `2012-01-01T01:01:01.000Z` is still acceptable, we probably need to make this guy a regexp to accept all kinds of ISO8601 compatible serialization.

pakrym: Ah, I didn't notice I've changed the value. Will reopen.

PR #30, opened 2020-04-07
must match (regexp) constraint

khenidak:
Implements must MustMatch<RegExp> constraint. Due to the way typescript process string (when they are a generic type constraint), **user must double escape the regexp**.

Use it:
```typescript
interface Person{
FirstName: string & MustMatch<'DOUBLE-ESCAPED-REG-EXP'>;
}
```

You can also define a custom type
```typescript
type email = string & MustMatch<'email-double-escaped-reg-exp'>;
```
then you can use it  as 
```typescript
interface person{
Email: email; // using defined type with its own constraint
}
```

### DIFF
--- a/adl.runtime/machinery/constraints/validation.ts
+++ b/adl.runtime/machinery/constraints/validation.ts
@@ -2,6 +2,36 @@ import * as adltypes from '@azure-tools/adl.types'
 import * as machinerytypes from '../machinery.types'
 import * as modeltypes from '../../model/module'
 
+export class MustMatchImpl implements machinerytypes.ValidationConstraintImpl{
+    Run(
+        context: machinerytypes.ConstraintExecContext,
+        rootTyped: any,
+        leveledTyped: any,
+        existingRootTyped: any | undefined,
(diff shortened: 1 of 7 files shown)


adl.runtime/machinery/constraints/validation.ts
```diff
@@ -2,6 +2,36 @@ import * as adltypes from '@azure-tools/adl.types'
...
+       const propVal = leveledTyped[context.propertyName];
+       let regExp = context.ConstraintArgs[0];
+       regExp = regExp.replace(/\\\\/g, '\\');
+       const re = new RegExp(regExp);
+       context.opts.logger.verbose(`constraint: MustMatch prop:${context.propertyName} with value (${propVal}) applying compiled regExp:${re.source}`);
+
+       if(!re.test(propVal)){
+         const propModel = leveledApiTypeModel.getProperty(context.propertyName) as modeltypes.ApiTypePropertyModel;
+         let details = " ";
+         if(propModel.isAliasDataType){
```
sanjaiganesh, line 25: what does alias data type mean?

PR #43, opened 2020-04-15
[Printer Part 1] Refactor printer

hjhhh3000vitae:
First pull request for table printer.

In this PR, `cmd_show_store` is refactored so that it could use a `--output=text` parameter to choose which printer to use.

### DIFF
--- a/cli/src/cmd_showstore.ts
+++ b/cli/src/cmd_showstore.ts
@@ -30,16 +30,18 @@ export class showStoreAction extends CommandLineAction {
     private _filter: CommandLineStringParameter; /* TODO */
     private _show_docs: CommandLineFlagParameter;
     
-    private readonly _output_format: string;
+    private _output_format: string;
 
-    public constructor(private ctx: appContext, outputFormat: string) {
+    public constructor(private ctx: appContext) {
     super({
       actionName: 'list',
       summary: 'shows the apis already loaded in the store',
       documentation: ''
     });
+  }
 
-    this._output_format = outputFormat;
+  public setOutputFormat(outputFormat: string) {
+      this._output_format = outputFormat;
   }
 
   private getPrinter(): printer {
--- a/cli/src/cmd_showstore.ts
+++ b/cli/src/cmd_showstore.ts
@@ -2,6 +2,8 @@ import { CommandLineAction, CommandLineChoiceParameter, CommandLineFlagParameter
 import { appContext } from './appContext'
 
 import * as adlruntime from '@azure-tools/adl.runtime'
+import { printer } from './printers/printer';
+import { textPrinter } from './printers/text_printer';
 
 /* shows what is in store
  * in a runtime env, this will connect to rpaas api server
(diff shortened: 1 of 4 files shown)


cli/src/commands.ts
```diff
@@ -75,6 +84,8 @@ export class adlCliParser extends CommandLineParser {
         // use it to create an api manager (store)
         ctx.store = machinery.createApiManager();
 
+        // setup output format
+        (<showStoreAction>this.getAction("list")).setOutputFormat(this._output_format.value ?? "text");
```
khenidak, line 88: i think we will have to redo that for every action, because every action uses the printers. So a better way is to set it on `AppContext` which is already wired down to each action.

cli/src/printers/printer.ts
```diff
@@ -0,0 +1,9 @@
+import * as adlruntime from '@azure-tools/adl.runtime'
+
+export interface printer {
+  printModel(model: adlruntime.ApiModel): void;
+  printNormalizedTypes(normalizedTypes: Iterable<adlruntime.NormalizedApiTypeModel>): void;
+  printApiVersions(apiVersions: Iterable<adlruntime.ApiVersionModel>): void;
+  printDocs(docs: adlruntime.ApiJsDoc | undefined): void;
+  flushOutput(): void;
```
khenidak, line 8: Why do we need this?

cli/src/cmd_showstore.ts
```diff
@@ -158,12 +64,15 @@ export class showStoreAction extends CommandLineAction {
...
                 // dumb print information from store as is
                 var models = this.ctx.store.ApiModels;
+
+                var printer = this.getPrinter();
                 // api infos
                 for(let model of models){
-                        this.printModel(scope, model);
-                        this.printNormalizedTypes(scope, model.NormalizedTypes);
-                        this.printApiVersions(scope, model.Versions);
+                        printer.printModel(model);
```
khenidak, line 71: Should be `printers.PrintModel(some filtering/options)` That way the caller does not to feed the printer everything. And printers can make independ choices about printing.

cli/src/cmd_showstore.ts
```diff
@@ -150,6 +40,22 @@ export class showStoreAction extends CommandLineAction {
     });
   }
 
+  public setOutputFormat(outputFormat: string) {
```
khenidak, line 43: Once we move to setting at AppContext (as discussed here: https://github.com/Azure/adl/pull/43/files#r410445314), we will no longer need this

hjhhh3000vitae:
As suggested, let's cancel this PR and move to https://github.com/Azure/adl/pull/44

cli/src/commands.ts
hjhhh3000vitae: Thx for the suggestion. Made it into `AppContext` (in Part 2)

cli/src/cmd_showstore.ts
hjhhh3000vitae: Yep. Removed.

cli/src/cmd_showstore.ts
hjhhh3000vitae: Updated accordingly.

cli/src/printers/printer.ts
hjhhh3000vitae: It is for `tablePrinter` only. `tablePrinter` won't know how wide it should reserve for one column until it processed all the data. Thus it cannot do it in a `print out while processing` fashion ---> need `flushOutput` to signal all data processed / do the actual printer.

PR #44, opened 2020-04-16
[Printer Part 1+2] table printer

hjhhh3000vitae:
Implemented table printer.

We can have https://github.com/Azure/adl/pull/43 merged then this one to keep things separate.

### DIFF
--- a/cli/src/cmd_showstore.ts
+++ b/cli/src/cmd_showstore.ts
@@ -4,6 +4,7 @@ import { appContext } from './appContext'
 import * as adlruntime from '@azure-tools/adl.runtime'
 import { printer } from './printers/printer';
 import { textPrinter } from './printers/text_printer';
+import { tablePrinter } from './printers/table_printer';
 
 /* shows what is in store
  * in a runtime env, this will connect to rpaas api server
@@ -46,11 +47,12 @@ export class showStoreAction extends CommandLineAction {
       switch (this._output_format) {
           case 'text': {
               return new textPrinter(this._scope.value ?? 'all', this._show_docs.value);
-              break;
+          }
+          case 'table': {
+              return new tablePrinter(this._scope.value ?? 'all', this._show_docs.value);
           }
           default: {
               return new textPrinter(this._scope.value ?? 'all', this._show_docs.value);
-              break;
           }
       }
   }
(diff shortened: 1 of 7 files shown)


cli/src/cmd_showstore.ts
```diff
@@ -27,120 +30,8 @@ export class showStoreAction extends CommandLineAction {
     private _scope: CommandLineChoiceParameter;
     private _filter: CommandLineStringParameter; /* TODO */
     private _show_docs: CommandLineFlagParameter;
-
```
khenidak, line 30: I think if you rebased this PR on top of `Part1` you will need to do the change again.

cli/src/printers/table_printer.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    let outputLine = "";
+    for (const [index, key, value] of header) {
+      if (value.length == 0) {
+        outputLine += "null";
+        outputLine += " ".repeat(columnWidth[index]);
+      }
+      else {
+        outputLine += value;
+        outputLine += " ".repeat(columnWidth[index] + 4 - value.length);
```
khenidak, line 86: a helper function that would return the padding string would be a lot more readable.

cli/src/printers/table_printer.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      if(prop.DataTypeKind == adlruntime.PropertyDataTypeKind.Complex ||
+        prop.DataTypeKind == adlruntime.PropertyDataTypeKind.ComplexArray ||
+        prop.DataTypeKind == adlruntime.PropertyDataTypeKind.ComplexMap){
+            this.printApiTypeModel(version, type, prop.getComplexDataTypeOrThrow(), propertyPrefix + prop.Name + '.');
+      }
+    }
+  }
+
+  private getPropertiesConstraintsAsText(p: adlruntime.ApiTypePropertyModel): string {
```
khenidak, line 133: if both printers are using this function then let us move it to a util/helper file and call it from both locations.

cli/src/cmd_showstore.ts
hjhhh3000vitae: Yep. I'll cancel Part 1 and focus on Part 2.

cli/src/printers/table_printer.ts
hjhhh3000vitae: Since we do not include `constraints` in table printer now. I'll remove it from this printer. Thx.

cli/src/printers/table_printer.ts
hjhhh3000vitae: Done so. Thanks.

PR #52, opened 2020-04-28
Load more primitive schemas from OpenAPI 3

nguerrera:


adl/core/serialization/openapiv3/schema.ts
```diff
@@ -261,18 +260,47 @@ export async function processIntegerSchema($: Context<v3.Schema>): Promise<Schem
...
       break;
+
+    default:
+      throw new Error(`Unexpected integer format: ${format}`);
+  }
+
+  return constrainNumericSchema($, result);
+}
+
+export async function processNumberSchema($: Context<v3.Schema>): Promise<Schema | undefined> {
```
fearthecowboy, line 271: good idea

nguerrera:
Rebased and adapted to new tracking scheme. Ready for review. The third commit is new since prior sign-off. The other two were adjusted during rebase to resolve conflicts.

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -197,55 +197,66 @@ export async function processStringSchema(schema: v3.Schema, $: Context): Promis
 }
 
 export async function processByteArraySchema(schema: v3.Schema, $: Context): Promise<Schema | undefined> {
-  return undefined;
+  use(schema.type);
```
fearthecowboy, line 200: I wonder if we should check for leftover unused properties here (in the event that they put something unexpected in).

On the other hand, if we check the attic when we're done, we'll see if we're not handling a case (but then we are relying on the tests to find everything.)

Hmm. I could put the check into the process function -- I'm already making it mark the source object as used, (but doesn't do it recursively) -- I could then just make it check to see if the whole object was used at that point. 

That's probably better since we don't need to write the same check over and over again

fearthecowboy: I tried adding a check in at the end of `context.process`  -- the trouble is that I've added in enough code underneath to remove almost all of the `use(...)`  calls, it will automatically call `use` when the objects are added into the target graph. 

But... that happens a lot later. 

I think checking the attic will be better, since it can be done once at the end.

PR #57, opened 2020-04-30
Load enums from OpenAPI v3

nguerrera:


adl/core/serialization/openapiv3/schema.ts
```diff
@@ -46,11 +46,10 @@ export async function processSchemaReference(ref: JsonReference<v3.Schema>, $: C
...
 
-
-export async function processSchema(schema: v3.Schema, $: Context, isAnonymous = false): Promise<Schema | undefined> {
-
-  // is enum or x-ms-enum specified?
-  if (schema.enum || (<any>schema)['x-ms-enum']) {
+export async function processSchema(schema: v3.Schema, $: Context, options?: { isAnonymous?: boolean, forUnderlyingEnumType?: boolean }): Promise<Schema | undefined> {
+  // if enum or x-ms-enum is specified, process as enum
+  // but not if we're already processing the enum and are now processing its underlying type
+  if (!options?.forUnderlyingEnumType && (schema.enum || (<any>schema)['x-ms-enum'])) {
```
fearthecowboy, line 52: Whoa. 

Yeah, that's an interesting notion. 

Historically, we've 'assumed' that `enum` was `string` (which was terrrrrrrrible ) and this has caused us a lot of grief when suddenly someone assumed they could do what they wanted.

I wonder if JSON Schema allows anything other than primitives... [Yep](http://json-schema.org/draft/2019-09/json-schema-validation.html#enum)... _Elements in the array might be of any type, including null._ 

Well, that sucks. 

Which means we have to be able to model some twisted weird enum value or constant that can be an object.

adl/core/serialization/openapiv3/header.ts
```diff
@@ -43,14 +43,13 @@ export async function processHeaders(input: ItemsOf<v3.Header>, $: Context): Pro
   return undefined;
 }
 
-async function processHeaderReference(headerReference: v3.HeaderReference, $: Context, isAnonymous = false): Promise<Header | undefined> {
-
+async function processHeaderReference(headerReference: v3.HeaderReference, $: Context, options?: { isAnonymous?: boolean }): Promise<Header | undefined> {
```
fearthecowboy, line 46: `options` -- interesting notion. 

Are there other `options` that you're thinking of? Oh. I see.

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -364,16 +363,50 @@ export async function processFileSchema(schema: v3.Schema, $: Context): Promise<
...
+    name?: string
+  };
+
+  type EnumInfo = {
+    name?: string,
+    modelAsString?: boolean,
+    values: EnumValue[]
+  };
+
+  // not using $.process here because we need to process a node that is already marked
```
fearthecowboy, line 379: 👍

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -364,16 +363,50 @@ export async function processFileSchema(schema: v3.Schema, $: Context): Promise<
...
+  const type = await processSchema(schema, $, { forUnderlyingEnumType: true }) ?? $.api.schemas.Any;
+  let enumValues = use(schema.enum);
+  let enumInfo = <EnumInfo>use((<any>schema)['x-ms-enum']);
```
fearthecowboy, line 382: It occurs to me that we should add the 'well known' ` x-ms-*` things to the OpenAPI model definitions so we can use them without having to resort to `<any>` all over the place

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -364,16 +363,50 @@ export async function processFileSchema(schema: v3.Schema, $: Context): Promise<
...
+  }
+
+  const result = new Enum(type);
+  result.modelAsString = enumInfo.modelAsString ?? false;
+  result.name = enumInfo.name;
 
-  if (length(schema.enum) === 0 && length(xmsEnum) === 0) {
-    // an enum with no values?
+  for (const each of enumInfo.values) {
+    const constant = new Constant(type, use(each.value));
```
fearthecowboy, line 396: GOOD IDEA!

I was just about to ask if we should make enum values themselves same thing as constants.

adl/core/serialization/openapiv3/schema.ts
fearthecowboy: Yeah, let's go add the known `x-ms-*` types directly into the OpenAPI models we've got -- we'll get strong type info out of it all, and it'll reduce duplication up front. 

Any types/interfaces should probably be added to common and  referenced in v2/v3

adl/core/serialization/openapiv3/schema.ts
nguerrera: Yes, this is weird, and I was hoping the answer would be that I could do something simpler...

adl/core/serialization/openapiv3/header.ts
nguerrera: Yeah, I wasn't sure about this. My main goal here was passing that extra bool to ProcessSchema without having to make other layers know about it unnecessarily. Secondarily, I also liked the side effect that you see boolean names at call site. In C#, I like to use named arguments whenever there's a bool.

adl/core/serialization/openapiv3/schema.ts
nguerrera: Done.

adl/core/support/visitor.ts
```diff
@@ -163,11 +164,7 @@ export class Context<TSourceModel extends OAIModel> {
     return this.visitor.api;
   }
 
-  async processAnonymous<TInput, TOutput extends Element>(action: fnAction<TSourceModel, TInput, TOutput>, value: TInput | NonNullable<TInput>): Promise<TOutput | undefined> {
-    throw undefined;
-  }
```
nguerrera, line 168: I'm not really sure what the intent of this function was. Should I have kept it?

fearthecowboy: No, we can ditch that. I wasn't sure if process was going to work with anonymous things or not, and now we know it's ok.

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -46,14 +46,16 @@ export async function processSchemaReference(ref: JsonReference<v3.Schema>, $: C
...
 
-
-export async function processSchema(schema: v3.Schema, $: Context, isAnonymous = false): Promise<Schema | undefined> {
-
-  // is enum or x-ms-enum specified?
-  if (schema.enum || (<any>schema)['x-ms-enum']) {
+export async function processSchema(schema: v3.Schema, $: Context, options?: { isAnonymous?: boolean; forUnderlyingEnumType?: boolean }): Promise<Schema | undefined> {
+  // if enum or x-ms-enum is specified, process as enum
+  // but not if we're already processing the enum and are now processing its underlying type
+  if (!options?.forUnderlyingEnumType && (schema.enum || (<any>schema)['x-ms-enum'])) {
```
nguerrera, line 52: note to self: remove any here.

adl/core/serialization/openapiv3/schema.ts
```diff
@@ -328,23 +315,30 @@ function constrainNumericSchema(schema: v3.Schema, $: Context, target: Schema):
...
     return result;
   }
+
+  /*
+  if (!schema.minItems && !schema.maxItems && !schema.uniqueItems) {
+    $.api.schemas.primitives.push(result);
+    return result;
+
+  }
+*/
```
nguerrera, line 332: Can this be deleted?

fearthecowboy: ach. can't believe I'm not catching this stuff.

yeah

nguerrera:
Good to go?

fearthecowboy:
LGTM

PR #59, opened 2020-05-01
Align .gitattributes to eslint config in favor of UNIX line breaks

nguerrera:
This aligns .gitattributes eol=lf to .default-eslintrc.yaml and prevents getting tons of eslint errors if the repo is checked out with core.autocrlf set.


nguerrera:
For obvious reasons, the diff without whitespace changes shows the meaningful change to .gitattributes: 

https://github.com/Azure/adl/pull/59/files?diff=unified&w=1

PR #76, opened 2020-05-04
Load servers from OpenAPI v3

nguerrera:


adl/core/serialization/openapiv3/server.ts
```diff
@@ -1,5 +1,21 @@
...
+      description: value.description,
+      defaultValue: value.default,
+      allowedValues: value.enum
+    });
+
+    connection.variables.push(variable);
+  }
+
+  $.api.http.connections.push(connection);
+  return connection;
```
nguerrera, line 20: This mirrors processTags, but a few things that tripped me up in the process:

1. The naming is plural, but we process one at a time.
2. We pass the output array into processArray, but processArray doesn't push into output, and so we have to add it here.
3. We return something but nothing appears to use it.

adl/core/test/scenarios/v3/single/output/aws-app-insights.api.yaml
```diff
@@ -1,4 +1,71 @@
-http: {}
+http:
```
nguerrera, line 1: This one is missing a corresponind attic change and I can't figure out why

adl/core/serialization/openapiv3/server.ts
nguerrera: My intuitive expectation with processArray taking an input -> output function and an output array parameter would be the one to do the pushing. Do you think that would be better? If so, I can refactor this and make processTags match.

fearthecowboy: I'm a bit confused as to why I wouldn't have made `processArray ` push to the target array. 

Oh. I see...  I wanted to make sure the collection got tracked at the end. It really never occured to me that the items should be pushed to the array. That sounds better, yeah.

I did refactor how this worked about 5 times while I was chasing a bug in the source tracking; I think I stopped when it was giving me the results I wanted. 😄

adl/core/model/http/protocol.ts
```diff
@@ -7,7 +7,35 @@ export class Authentication extends Element {
...
 export class Connection extends Element {
+  variables = new ElementArray<ConnectionVariable>();
 
+  constructor(public url: string, initializer?: Partial<Connection>) {
+    super();
+    this.initialize(initializer);
+  }
+}
+
+// Consider modeling this as a Parameter once that's fleshed out?
```
fearthecowboy, line 18: > Consider modeling this as a Parameter once that's fleshed out?
Yeah, that would be a good idea.

adl/core/serialization/openapiv3/server.ts
```diff
@@ -1,5 +1,21 @@
...
-export async function processServers(): Promise<Element | undefined> {
-  return undefined;
+export async function processServers(server: v3.Server, $: Context) {
+  const connection = new Connection(server.url);
+
+  for (const { key, value } of items(server.variables)) {
+    const variable = new ConnectionVariable(key, {
+      description: value.description,
+      defaultValue: value.default,
+      allowedValues: value.enum
```
fearthecowboy, line 13: I wonder if we should synthesize an enum for `allowedValues` and use that. 

and then the default value could point to the enumvalue in the enum.  (or synthesize a constant  if there wasn't an enum)

That might make the transition to a parameter pattern a bit simpler.

adl/core/test/scenarios/v3/single/output/aws-app-insights.api.yaml
fearthecowboy: O_o Like ... the attic didn't change at all? 

Sounds like one more bug in source tracking -- I'll debug that when this is committed.

adl/core/serialization/openapiv3/server.ts
```diff
@@ -1,5 +1,21 @@
...
+import { Connection, ConnectionVariable } from '../../model/http/protocol';
+import { Context } from './serializer';
+import { items } from '@azure-tools/linq';
 
-export async function processServers(): Promise<Element | undefined> {
-  return undefined;
+export async function processServers(server: v3.Server, $: Context) {
+  const connection = new Connection(server.url);
+
+  for (const { key, value } of items(server.variables)) {
```
fearthecowboy, line 9: > `items(server.variables)`
probably need to say `items(use(server.variables))` -- if we don't use `process` on the collection it doesn't automatically 'use' the collection.

fearthecowboy: This might be why it's not clearing the attic.

nguerrera: This was part of it. I was missing server.description also and not `use()`ing server.enum. I also couldn't get items to work so did a for...in. 

I found this quite hard to debug. I need to study the tracking system to gain a deeper understanding. The handling of children and parents and what not is mysterious to me right now as I'm trying to step through it. The challenge right now is that the attic was all or nothing until I added all of the missing changes. So it didn't tell me anything other than "something is wrong in processServers".

adl/core/serialization/openapiv3/server.ts
nguerrera: I think this would be smart. The order I think would make sense would be:

1. Start with this as is for servers
2. Fully model parameters
3. Go back and make these variables into parameters of type string or enum-of-string.

What do you think?

adl/core/serialization/openapiv3/server.ts
nguerrera: Figured out how to use `items`, needed a use on the value. Fixed.

adl/core/model/http/protocol.ts
nguerrera: See https://github.com/Azure/adl/pull/76#discussion_r420306662

fearthecowboy:
LGTM!

PR #80, opened 2020-05-06
Fix eslint issues and run eslint on PRs

nguerrera:


fearthecowboy:
Hmm. I had fix on save on. I wonder why it stopped working...

fearthecowboy:
Ah. Missing some configuration in .vscode 

Fixing and will push an update.

PR #86, opened 2020-05-06
Components (parts we need) Complete

fearthecowboy:
- disabled VSCode's format on save, eslint is doing it now (and vscode actually makes eslint mad)
- generic alias type for things that are not schemas 
- refactored ref handling a bit more to make it very generic


adl/core/serialization/openapi/v3/schema.ts
```diff
@@ -79,8 +64,25 @@ function commonProperties(schema: v3.Schema) {
   };
 }
 
-export async function processInline(schema: v3.Schema | v3.SchemaReference | undefined, $: Context, options?: Options) {
-  return schema ? $.processPossibleReference(processSchemaReference, processSchema, use(schema), { ...options, isAnonymous: true }) : undefined;
+//export async function processInline(schema: v3.Schema | v3.SchemaReference | undefined, $: Context, options?: Options) {
+//  return schema ? $.processPossibleReference(processSchemaReference, processSchema, use(schema), { ...options, isAnonymous: true }) : undefined;
+// }
```
nguerrera: Delete?

PR #91, opened 2020-05-08
 Refactor schema

fearthecowboy:
- pulled in linq project to make perf changes 
  - `values`, `keys`, `items` return straight Iterable<> for faster use
  - **important** -- `items()` now returns an array [key,value] instead of an object {key, value} -- change usage from `for( const {key,value} of items(foo) (` to  `for ( const [key, value] of items(foo) ) `
  - switched to using `Object.entries()` and `Object.values()` for object/dictionaries 
  - introduced `linq.values`, `linq.keys`, `linq.items` when you want a linq operation on the iterable (ie, returns a `IterableWithLinq<>`
- schemas are now handled with iterators, no direct pushing to the target collection
- aliases are created when there is a description/nullable on a primitive type
-


adl/core/serialization/openapi/v3/components.ts
```diff
@@ -1,52 +1,84 @@
 import { v3 } from '@azure-tools/openapi';
+import { Alias as A } from '../../../model/alias';
```
nguerrera, line 2: Maybe a more descriptive disambiguating name than 'A'?

adl/core/test/scenarios/v3/single/output/github.attic.yaml
```diff
@@ -1 +1,146 @@
-{}
+paths:
```
nguerrera, line 1: stuff is going *into* the attic?

fearthecowboy: WTF  it doesn't look like that locally. 

One sec.

fearthecowboy: Ok, I'm confused -- I pushed it, and it was correct in my tree, but the PR didn't pick it up for a while. I've never seen that happen

PR #93, opened 2020-05-08
whitespace changes

fearthecowboy:


documentation/examples/prototype/schemas/thirdProperties.adl.ts
```diff
@@ -1,4 +1,5 @@
 import { Color } from '../enums/color.adl';
+;
```
nguerrera, line 2: Wonder what happened here

PR #95, opened 2020-05-11
Paths and schemas v2

fearthecowboy:
- merged security/etc from master
- added fix to not loop forever when circular references are added to the model
- added v2 schema processing -- made as much reusable as possible
- added all the x-ms-* extensions to the oai defs -- most of this goes to the element attic for now, since it's directed to client generation
- paths are handled -- operations with a parameter of type body will synthesize a `Request`  for them


adl/sourcemap/proxies.ts
```diff
@@ -337,7 +337,11 @@ export class TrackedTarget<T extends Object> {
     return property in actual;
   }
 
-  onAdd(tracker: Tracker, pathInTarget: Path) {
+  onAdd(tracker: Tracker, pathInTarget: Path, cache = new WeakSet()) {
+    if( cache.has(this.proxy)) {
```
fearthecowboy, line 341: this prevents never-ending calls when a circular reference is added to the model.

PR #98, opened 2020-05-14
Add tracking

fearthecowboy:
- updated to latest TS release  - 3.9.2
- adds a trackmap/sourcemap pattern into element
- adds a means to get a 'path' to a node so it can be found again (happens if the node instance gets invalidated because of source edits)
- added a simple proxy for nodes to do an inline lookup of the node if it gets invalidated


adl/core/model/schema.ts
```diff
@@ -163,17 +183,21 @@ class EnumActual extends Schema implements Enum {
...
+  addValue(name: string, value: string | number): EnumValue {  
     const em = this.node.addMember({
       name: quoteForIdentifier( valueOf(name)),
       value: valueOf(value),
     });
-    
-    const ev = new EnumValue(em);
 
+    const ev = new EnumValue(em).track({
+      // he who creates a trackable object shall call track
```
fearthecowboy, line 196: pretty much every time we create a node in TS, we have to tell it where the values came from

adl/core/model/schema.ts
```diff
@@ -153,6 +167,12 @@ class InlineEnum extends Schema implements Enum{
 
 class EnumActual extends Schema implements Enum {
   node: EnumDeclaration;
+  get targetMap(): Dictionary<any> {
```
fearthecowboy, line 170: for every `Element` that has a representation in a node, it should implement `targetMap` accessor that gets the `Path` for each thing we can track to something in the original OAI YAML  files

adl/core/model/schema.ts
```diff
@@ -163,17 +183,21 @@ class EnumActual extends Schema implements Enum {
   
   constructor(decl: EnumDeclaration) {
     super('enum');
-    this.node = decl;
+    this.node = referenceTo(decl);
```
fearthecowboy, line 186: ### Important
We have to wrap the node using `referenceTo` -- the node itself is fairly ephemeral, and can get invalidated if the TS file that it's in gets edited. 

`referenceTo` creates a simple proxy that checks to see if the node was invalidated, and if it was it uses the `Path` to find it again.

adl/core/model/schema.ts
```diff
@@ -197,28 +221,32 @@ export function createEnum(api: ApiModel, elementSchema: Schema, initializer?: {
...
-
+      
       const file = api.getEnumFile(name);
+      
       const result = new EnumActual(file.addEnum({
-        name,
+        name: valueOf(name),
         isExported: true,
-      }));
+      })).track({
```
fearthecowboy, line 242: using `track` will also mark the value in the source YAML file as 'used'

adl/core/support/typescript.ts
```diff
@@ -0,0 +1,76 @@
...
+    (<any>node).getFilePath ? (<any>node).getFilePath() : 
+      (<any>node).getValue ? (<any>node).getValue() : 
+        node.getChildIndex(); 
+}
+/**
+ * returns a Path to the node that we can use to find it again.
+ * 
+ * @param node the node to create the path for.
+ */
+export function getPath(node: Node, ...args: Path ): Path{
```
fearthecowboy, line 22: This builds a 'path' to the target node by finding the `name` or `filePath` or `value` (and failing all that, the `index` in the parent)  of each step in the ancestor chain, so that when we have to find the node again, we can just follow the breadcrumbs :D

fearthecowboy: I'm sure it's not perfect, but it should suffice to keep track of things for the immediate timeframe

adl/core/support/typescript.ts
```diff
@@ -0,0 +1,76 @@
...
+    has: (originalTarget: T, property: keyof T): boolean => {
+      return Reflect.has(current.wasForgotten() ? (current = <T>getNode([...path], project)) : current, property);
+    },
+    set: (originalTarget: T, property: keyof T, value: any, receiver: any)=> {
+      return Reflect.set(current.wasForgotten() ? (current = <T>getNode([...path], project)) : current, property, value);
+    }
+  });
+}
+
+export function project<T extends Node>(input: T): ApiModel {
```
fearthecowboy, line 74: This lets us get a reference back to the `ApiModel` from any node that's in any document in the project

adl/core/support/typescript.ts
```diff
@@ -0,0 +1,76 @@
...
+  return result && path.length ? getNode( path, result) : result;
+}
+
+/**
+ * Creates a reference to the node that will reacquire the target if the AST forces the node object to be forgotten or invalid.
+ * 
+ * @param input the node to create a reference for.
+ */
+export function referenceTo<T extends Node>(input: T): T { 
+  const project = input.getProject();
```
fearthecowboy, line 54: creates a simple proxy that can switch it's target if it gets invalidated, by using the path to get back to the correct node

adl/core/model/schema.ts
```diff
@@ -1,13 +1,21 @@
...
+import { getPath, referenceTo } from '../support/typescript';
 import { ApiModel } from './api-model';
 import { Element } from './element';
 import { Identity } from './name';
 
-
 export class Schema extends Element {
+  get targetMap(): Dictionary<any> {
+    return {
+      ...super.targetMap,
```
fearthecowboy, line 13: it's not as automatic as the original proxies were, but it's about as simple as I could make it.

PR #100, opened 2020-05-15
Flesh out enum -> ADL

nguerrera:


adl/core/test/scenarios/v2/multiple/apimanagement/output/ContentFormat.ts
```diff
@@ -1,4 +1,6 @@
-/** @description Format of the Content in which the API is getting imported.
+/**
```
nguerrera, line 1: Complete example here.

adl/core/model/schema/enum.ts
```diff
@@ -29,115 +39,124 @@ export class EnumValue extends Element {
...
+}
 
-      const file = api.getEnumFile(name);
+export function createEnum(api: ApiModel, values: Array<EnumValue>, initializer?: Partial<Enum>) {
+  const name = initializer?.name ?? anonymous('enum');
 
-      const result = new EnumActual(file.addEnum({
-        name: valueOf(name),
-        isExported: true,
-      })).track({
```
nguerrera, line 126: didn't mean to lose the track calls, it was a casualty of merging across different files.

adl/core/model/element.ts
```diff
@@ -92,4 +97,83 @@ export class Element extends Initializer {
...
+
+    if (!Node.isJSDocableNode(this.node)) {
+      throw new Error('This node cannot have JS documentation');
+    }
+
+    return this.node.addJsDoc({});
+  }
+
+  protected hasDocTag(tagName: string) {
+    return this.getDocTag(tagName) !== undefined;
```
fearthecowboy, line 127: How will this work for doctags where there are multiple of them (ie, `@param` )?

adl/core/model/element.ts
```diff
@@ -92,4 +97,83 @@ export class Element extends Initializer {
...
+
+  protected removeTag(tagName: string) {
+    for (const each of this.getDoc()?.getTags() ?? []) {
+      if (each.getTagName() == tagName) {
+        each.remove();
+      }
+    }
+  }
+
+  protected getDocSummary() {
```
fearthecowboy, line 169: should we just make these just an accessor pair for `summary`?

adl/core/model/schema/enum.ts
```diff
@@ -29,115 +39,124 @@ export class EnumValue extends Element {
...
     return {
       ...super.targetMap,
       $: getPath(this.node),
     };
   }
 
-  get sealed() {
-    return true;
+  get extensible() {
+    return this.hasDocTag('extensible');
```
fearthecowboy, line 87: what if someone edits the doc and literally says `@extensible false` ?

fearthecowboy: oh, it's a marker, not a value. my bad.

adl/core/model/element.ts
```diff
@@ -92,4 +97,83 @@ export class Element extends Initializer {
...
+        if (each) { // strangely we can get undefined docs back
+          return each;
+        }
+      }
+    }
+
+    return undefined;
+  }
+
+  private getOrCreateDoc() {
```
fearthecowboy, line 113: can we make this a property getter?

fearthecowboy: and shouldn't this be protected to allow children to create their own tags for things?

fearthecowboy:
I think the whitespace problems are totally my fault.

adl/core/model/element.ts
nguerrera: Yeah, the whole doc convenience API probably needs a lot more work. 

Here, getDocTag returns the first one this one just returns a bool, so if you used getDocTag('param') it would tell you there's at least one param, but this wasn't really thought out for that case. I can imagine ways to have different convenience APIs for things like param.

adl/core/model/element.ts
nguerrera: It could be protected. I was trying to guide subclasses to use the helpers built on this, and not this directly. But as you pointed out above, the helpers aren't up to everything we'll need.

nguerrera:
Merging to unblock parallel work on  this as a baseline, but will adress feedback in subsequent PR.

PR #106, opened 2020-05-21
Align enums to recent object changes

nguerrera:


adl/test/scenarios/v3/single/output/twitter/models/Poll.ts
```diff
@@ -7,5 +7,5 @@ export interface Poll {
     end_datetime: dateTime;
     id?: string & RegularExpression<"^[0-9]{1,19}$">;
     options?: Array<PollOption> & MaximumElements<4> & MinimumElements<2>;
-    voting_status: unknown /*= (not tsschema -- undefined[object Object]/undefined ) =*/;
+    voting_status: "open" | "closed";
```
nguerrera, line 10: Example of desired change among all the renumbering :)

adl/core/model/schema/enum.ts
```diff
@@ -90,73 +63,89 @@ class GenuineEnum extends TSSchema<EnumDeclaration> implements Enum {
...
+  constructor(node: EnumDeclaration, private anonymous: boolean) {
+    super('enum', node);
+    this.values = new CollectionImpl(this, this.addValue, this.removeValue, this.getValues);
+  }
 }
 
-export function createEnum(api: ApiModel, values: Array<EnumValue>, initializer?: Partial<Enum>) {
-  const name = initializer?.name ?? anonymous('enum');
+export function createEnum(api: ApiModel, identity: Identity, values: Array<EnumValue>, initializer?: Partial<Enum>) {
+  const { name, file } = api.getNameAndFile(identity, 'model');
```
fearthecowboy, line 122: should be 'enum'

fearthecowboy: this causes `model_311` in a test below.

fearthecowboy, line 122: test

adl/core/model/element.ts
```diff
@@ -45,7 +45,6 @@ export class Initializer {
         proxy[key] = (<any>value);
       }
     }
```
fearthecowboy, line 47: uh, deleted a line?

adl/core/model/schema/enum.ts
nguerrera: Sigh I managed to fix this without pushing it. I blame ready-for-lunch-brain :).

adl/core/model/element.ts
nguerrera: The return this was causing a problem when I had a T1 | T2 and called initialize on it. But I don't have that anymore so reverting.

nguerrera: (We weren't using the return value AFAICT.)

nguerrera: Reverted

adl/core/model/schema/enum.ts
nguerrera: Fixed

PR #111, opened 2020-05-21
speed up tests!

fearthecowboy:


adl/core/model/api-model.ts
```diff
@@ -242,12 +185,12 @@ export class ApiModel {
 
     reqdTypes:
     for (const requiredType of schema.requiredTypeDeclarations) {
-
+
```
nguerrera, line 188: nit: whitespace (testing commenting from VS Code)

PR #116, opened 2020-05-29
Disable source tracking, remove target tracking entirely

fearthecowboy:
This PR: 
- gets rid of all the target node tracking
- and removes the source tracking on all non-object nodes (we still need it on object nodes to know the current location/name of the current node)

I removed all the `use(...)` calls, and the `valueOf(...)` calls, since they are never needed (because all primitive/string values are always the unboxed value now)

This also means we have no `isUsed` functionality for now, nor any ability to know what didn't get processed. 

After we have all the processing done, I'll revisit how to track usage and create source map information for the nodes (since I already have the plumbing and know that I can do it with the proxy, it's just a matter of inserting the functionality in without breaking the test outputs.)

### DIFF
--- a/adl/core/.eslintignore
+++ b/adl/core/.eslintignore
@@ -1,2 +1,3 @@
 **/*.d.ts
 test/scenarios/**
+yarn/*.js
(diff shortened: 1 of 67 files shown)


nguerrera:
Thanks! .

nguerrera:
Good to merge?

PR #118, opened 2020-05-29
Start emitting operations

nguerrera:
This is incomplete, but a reasonable checkpoint.

There are places where we can load things, but not edit or read them back. There's still more to load/emit too. Probably most importantly, requests and responses aren't in the codegen. I will look at that next, then flesh out the other details.


adl/test/scenarios/v3/single/output/ablyio/operations/Service.ts
```diff
@@ -0,0 +1,178 @@
+export interface Service {
+    /**
+     * Enumerate all active channels of the application
+     * @description Enumerate all active channels of the application
+     * @http GET /channels
+     * @tag Status
+     * @since 1.1.0
+     */
+    getMetadataOfAllChannels(_0?: Http.Header<string, 'X-Ably-Version'>, _1?: Http.Query<"json" | "jsonp" | "msgpack" | "html", 'format'>, limit?: Http.Query<int64>, prefix?: Http.Query<string>, by?: Http.Query<"value" | "id">);
```
fearthecowboy, line 9: This is really cool stuff.

adl/core/model/operation.ts
```diff
@@ -1,33 +1,23 @@
...
+export interface Operation extends Element {
   /** A short summary of what the operation does. */
   summary?: string;
 
   /** A verbose explanation of the operation behavior. Commonmark syntax can be used for rich text representation. */
   description?: string;
 
   /** parameters common to all the requests(overloads) for this operation */
-  parameters = new Array<Parameter|Alias<Parameter>>();
+  readonly parameters: ReadOnlyCollection<Parameter | Alias<Parameter>>;
```
nguerrera, line 14: ReadOnlyCollection here it's incorrect to add a non-http parameter to an http operation. http subclass provides Collection<http.Parameter> for mutation.

adl/test/scenarios/v2/single/output/github/operations/Service.ts
```diff
@@ -0,0 +1,1637 @@
...
+     * @http GET /repos/{owner}/{repo}/commits/{shaCode}/comments
+     * @since v3
+     */
+    comments123456789(owner: Http.Path<string>, repo: Http.Path<string>, shaCode: Http.Path<string>, Accept?: Http.Header<string>);
+    /**
+     * @description Create a commit comment.
+     * @http POST /repos/{owner}/{repo}/commits/{shaCode}/comments
+     * @since v3
+     */
+    comments12345678910(owner: Http.Path<string>, repo: Http.Path<string>, shaCode: Http.Path<string>, Accept?: Http.Header<string>);
```
nguerrera, line 479: lol, the numbering here looks like a bug on my part.

Related: We're not succeeding at choosing a grouping for github and it makes ts-morph very slow on very big files. We're also not succeeding at coming up with a name. I thought about incorporating the http verb when there's a collision.

adl/core/model/http/operation.ts
```diff
@@ -1,46 +1,178 @@
...
+      appendTag(this.node, 'tag', each);
+    }
+  }
+    
+  private pushParameters(...parameters: Array<Parameter | Alias<Parameter>>) {
+    const structures = new Array<ParameterDeclarationStructure>();
+    for (let each of parameters) {
+      each = valueOf(each);
+      const parameter = each instanceof Alias ? each.target : each;
+      const name = normalizeName(each.name);
```
fearthecowboy, line 141: Hmm. We should be picking up the name of the parameter from the name alias when it is there.  (we're getting `_0`, `_1` etc.)

adl/test/scenarios/v2/single/output/github/operations/Service.ts
nguerrera: Fixed the numbering.

adl/core/model/http/operation.ts
nguerrera: Looking. weird because we are using each.name, not each.target.name in the alias case here. I'll debug.

adl/test/scenarios/v2/single/output/github/operations/Service.ts
fearthecowboy: Again, github comes thru with a bad enough example that we're going to have to dig deep and get smarter. 

Good thing we're getting hammered early with that. 

I'll get what I'm working on a bit more stable, and then I'll go update models to do the 'structure/create once' style creation and we can then apply the same thing in operations/parameters.

The hard part is models tend to have types that are recursive (ie `Error` has a collection of `Error`s), so if we go the 'create at once' style, it's really hard to get a reference to something that's not actually created yet.  

Maybe the solution to that is when things are `$ref`'d is to do some kind of shallow processing where we only do enough to get the declaration back, and let the definition get handled when the primary loop runs it. That way we'd be able to *use* something before we actually processed it. 

This may take me a bit of thinkin'

adl/core/model/http/operation.ts
fearthecowboy: Don't spend too much time on it -- if we go down the 'shallow' process on `$ref`s -- maybe this will be something that works itself out.

adl/test/scenarios/v3/single/output/twitter/operations/Service.ts
```diff
@@ -0,0 +1,76 @@
...
...me(username: Http.Path<schema>, _1?: Http.Query<Array<"pinned_tweet_id"> & MinimumElements<1> & UniqueElements, 'expansions'>, _2?: Http.Query<Array<"id" | "created_at" | "text" | "author_id" | "in_reply_to_user_id" | "referenced_tweets" | "attachments" | "withheld" | "geo" | "entities" | "public_metrics" | "possibly_sensitive" | "source" | "lang" | "context_annotations"> & MinimumElements<1> & UniqueElements, 'tweet.fields'>, _3?: Http.Query<Array<"id" | "created_at" | "name" | "username" | "protected" | "verified" | "withheld" | "profile_image_url" | "location" | "url" | "description" | "entities" | "pinned_tweet_id" | "public_metrics"> & MinimumElements<1> & UniqueElements, 'user.fields'>, _4?: Http.Query<Array<"media_key" | "duration_ms" | "height" | "preview_image_url" | "type" | "url" | "width"> & MinimumElements<1> & UniqueElements, 'media.fields'>, _5?: Http.Query<Array<"id" | "name" | "country_code" | "place_type" | "full_name" | "country" | "contained_within" | "geo"> & MinimumElements<1> & UniqueElements, 'place.fields'>, _6?: Http.Query<Array<"id" | "options" | "voting_status" | "end_datetime" | "duration_minutes"> & MinimumElements<1> & UniqueElements, 'poll.fields'>);
```
fearthecowboy, line 67: This API makes me all stabby 🗡️ 

My goodness, that's "impressive"

adl/core/model/http/operation.ts
nguerrera: It was a trivial change to fix it so I did.

PR #119, opened 2020-06-01
Emit requests and responses

nguerrera:


adl/core/model/http/operation.ts
```diff
@@ -176,4 +176,62 @@ class OperationImpl extends NamedElement<MethodSignature> implements Operation {
         throw new Error(`Invalid parameter type '${parameterType}'`);
     }
   }
-}
+
+  private pushRequests(...requests: Array<Request | Alias<Request>>) {
```
nguerrera, line 180: These are already capable of batching the insertion, but the caller would still need to be changed to stop calling push one at a time. The Request/Parameter/Response objects coming in are just data.

adl/core/model/http/operation.ts
```diff
@@ -176,4 +176,62 @@ class OperationImpl extends NamedElement<MethodSignature> implements Operation {
...
+    this.node.setReturnType(returnType);
+  }
+
+  private getResponseType(response: Response) {
+    const outerType = response.isException ? 'Http.Exception' : 'Http.Response';
+    const statusArg = this.getStatusArg(response);
+    const schema = response.schema ? this.project.getTypeReference(response.schema, this.node.getSourceFile()) : undefined;
+    const schemaArg = schema ? `, ${schema}` : response.mediaType ? ', none' : '';
+    const mediaTypeArg = response.mediaType ? `, '${response.mediaType}'` : '';
+    return `${outerType}<${statusArg}${schemaArg}${mediaTypeArg}>`;
```
nguerrera, line 230: This is messy. The string setType can get unwieldy. Emitting actual types somewhere temporary to get a string would be costly based on our experience with the API so far. There's probably at least some middle ground with some helper functions to build these strings. I'll give that a try.

PR #120, opened 2020-06-03
Include all responses and add request/response parameter descriptions

nguerrera:


nguerrera:
Perf continues to be an issue. Adding the doc comments really hurt it. I tried batching the low hanging fruit, but it didn't make a big dent :(. Held back from this review.

PR #135, opened 2020-06-12
added operations doc

fearthecowboy:

PR #137, opened 2020-06-14
 Align operation output to latest proposal

nguerrera:
Implements #135 

Note, however, this does NOT do the following yet:
- Make parameter, header, response type aliases. Everything is still inlined as before for now.
- Dedupe responses that differ only by code or mediaType into combined criterias with `|`s
- Put doc comments into response types. They are still `@return`, added mediaType to disambiguate.


adl/test/scenarios/v2/multiple/management-keyvault/output/operations/Vaults.ts
```diff
@@ -134,9 +156,11 @@ export interface Vaults {
      * @param __top - Maximum number of results to return.
      * @param api_version - Azure Resource Manager Api Version.
      * @param subscriptionId - Subscription credentials which uniquely identify Microsoft Azure subscription. The subscription ID forms part of the URI for every service call.
-     * @return 200 - Get information about all key vaults in the subscription.
+     * @return 200|application/json - Get information about all key vaults in the subscription.
```
fearthecowboy, line 159: I like this -- this is very helpful.

adl/core/serialization/openapi/v2/response.ts
```diff
@@ -12,11 +12,13 @@ export async function* response(response: v2.Response, $: Context, options?: { i
   const responseName = options?.isAnonymous ? anonymous('response') : nameOf(response);
   const operation = options?.operation;
   const produces = operation?.produces || $.sourceModel.produces || [];
+  const isException = !isAnonymous(responseName) && (responseName == 'default' || responseName.startsWith('4') || responseName.startsWith('5'));
```
fearthecowboy, line 15: Historical note: ` responseName.startsWith('4') ` -- we have RPs who have 4xx return codes that are *sigh*  not intended to be exceptional behavior.

(ie, 404 on a delete meant that the response was not an error, just 'ya, ok, it's gone')

I wonder if we should figure a way to put a `TODO: Assumption made - XXXX `comment in generated code when we are forced to make an interpretation, just so that people can find those when we do them.

What do you think?

adl/test/scenarios/v2/multiple/management-keyvault/output/operations/Vaults.ts
```diff
@@ -76,9 +90,11 @@ export interface Vaults {
      * @param __top - Maximum number of results to return.
      * @param api_version - Client Api Version.
      * @param subscriptionId - Subscription credentials which uniquely identify Microsoft Azure subscription. The subscription ID forms part of the URI for every service call.
-     * @return 200 - Get information about all key vaults in the specified resource group.
+     * @return 200|application/json - Get information about all key vaults in the specified resource group.
      */
-    ListByResourceGroup(resourceGroupName: Http.Path<string>, __top?: Http.Query<int32, '$top'>, api_version: Http.Query<string, 'api-version'>, subscriptionId: Http.Path<string>, body?: Http.Body<file, 'application/json'>): Http.Response<'200', [object, Object], 'application/json'>;
+    ListByResourceGroup(resourceGroupName: string, __top?: Query<int32, '$top'>, api_version: Query<string, 'api-version'>, subscriptionId: string, body?: Body<file, 'application/json'>): [(code: 200, mediaType: "application/json") => {
```
fearthecowboy, line 95: It does occur to me that `$` is a permissible character in identifiers -- is the `__` happening in a name formatter we can enhance? ideally, we don't adjust names if we don't absolutely have to.

fearthecowboy:
Sweet mercy this is looking nice.

adl/test/scenarios/v2/multiple/management-keyvault/output/operations/Vaults.ts
nguerrera: Yes, I think we just have our own overly conservative regex here. I started to use the compiler API here to overcome the fact that ts-morph wants a string for types, and it was much more pleasant then I expected. Also, seems fast enough. I imagine there might be compiler API we can leverage for creating identifiers too instead of having to work out regexes. I can look into that.

fearthecowboy: Really good point.

PR #138, opened 2020-06-15
Start Direction document

fearthecowboy:


documentation/Design/Direction.md
```diff
@@ -0,0 +1,141 @@
...
+}
+
+```
+
+> ADL V2 equivalent: 
+
+``` typescript 
+/**
+ * My Model type
+ * @since v1
```
MiYanni, line 117: Would these still be doc comments? Seems like, if you have full control, they would just be actual annotations. Meaning, `@since v1` above the declaration instead of in a comment block.

documentation/Design/Direction.md
```diff
@@ -0,0 +1,141 @@
...
+ */
+export model Person {  // <-- change the 'interface' to 'model', and remove the `@model` tag
```
MiYanni, line 119: Would we still need `export`? I'd think that in ADL, everything declared would be exported.

documentation/Design/Direction.md
```diff
@@ -0,0 +1,141 @@
...
+  name: string;
+
+  /** 
+   * Role the person has.
+   * @deleted v2
+   */
+  readonly role: 'Manager' | 'Employee';
+
+  /** 
+   * Role the person has
```
MiYanni, line 131: Similar to what I mentioned above about annotations, would be more consistent to no longer use comment blocks and have something like `@description Role the person has` instead.

documentation/Design/Direction.md
```diff
@@ -0,0 +1,141 @@
...
+
+    - package up the forked vscode extension, tooling and libraries in order to be able to bootstrap 
+      ADL consumers easily.
+
+  - `2.0` - __Deep fork__ - start enginieering the language to be exactly what we want to be.
+      - at some point in the not-too-distant future, we can opt to do extensive 
+      alteration to the fork of TS to allow us to enrich the language with more natural constructs 
+      instead of simply overloading the existing TS notation that we come up with.
+
+      - This enables us to richly extend the language in a forward compatible way when we're ready
```
MiYanni, line 46: The example of v2 below, where `role` is defined twice is not really 'forward compatible' with TS, right?

documentation/Design/Direction.md
```diff
@@ -0,0 +1,141 @@
...
+
+      - we can add linting and refactoring rules to auto-upgrade to the cleaner syntax to support
+      new constructs.
```
MiYanni, line 49: I gave some examples below of 'new constructs' that might make sense, like true `@` annotations (non-doc comments).

documentation/Design/Direction.md
fearthecowboy: Good Question.

It's possible we could support 'decorators' on things other than classes, which would let us annotate with those.

We could also use a different syntax.

documentation/Design/Direction.md
fearthecowboy: Ideally, yes. that's definitely something worth figuring out too.

documentation/Design/Direction.md
fearthecowboy: @description is literally a doc comment tag. I dunno that I'd change that just for fits and giggles.

documentation/Design/Direction.md
fearthecowboy: I meant that whatever we write now will still work in the future. I guess that's backward compatible in the future, but now, it's planning for it the other way 😸

documentation/Design/Direction.md
MiYanni: I meant that `@description` would be used in that 'actual annotations' strategy/concept. So, you wouldn't use comment blocks. I think that in ADL v2, comment blocks should *actually be comments*, not documentation for the API.

PR #139, opened 2020-06-17
Use type aliases for header definitions

nguerrera:
OK, after several false starts, I think I finally have a pattern that will work, but I've only managed to type-alias-ify headers with it. I will follow up with parameters, requests and responses tomorrow. Hopefully,  I can get through that relatively quickly now.

### Notes
* I renamed TypeDeclaration to TypeSyntax because there's another TypeDeclaration as union of ts-morph XxxDeclaration
* I made TypeSyntax a class that can either be a string or a compiler node, and you can ask it for a string or node and it will convert lazily. 
* The string -> node conversion currently cheats and uses a TypeReference with a "name" that is arbitrary syntax for a type like "a | b | c". 
* This happens to work and happens to be what I accidentally did the last time. I experimented with parsing the text to a proper tree node for this conversion, but it had some side effects on the output: doc comments inside anonymous types were dropped, single quotes became double quotes.
* It might be interesting to revisit that, and also interesting to consider using compiler nodes instead of strings in more places over time.

### Other minor changes in this PR
* set docs on type aliases
* remove empty docs
* use summary for nodes that cannot have summary and description

### Commits
52b67fe has all the code changes
63be550 isolates the meaningful output change to headers
3c951e0 has the rest of the output changes (docs)

### DIFF
--- a/adl/core/support/codegen.ts
+++ b/adl/core/support/codegen.ts
@@ -1,4 +1,4 @@
-import { ts } from 'ts-morph';
+import { printNode, ts } from 'ts-morph';
 
 /**  
  * Prepares a string for use as a TypeScript identifier.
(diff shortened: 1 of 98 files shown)


adl/core/support/codegen.ts
```diff
@@ -37,8 +37,42 @@ export function normalizeName(value: string ) {
...
+      // text here into a proper tree, but that costs extra parsing and has the
+      // side effect of losing trivia and formatting.
+      this.#node = ts.createTypeReferenceNode(this.#text!, undefined);
+    }
+    return this.#node;
+  }
+
+  get text() {
+    if (!this.#text) {
+      this.#text = printNode(this.#node!);
```
fearthecowboy, line 69: FYI - a very common javascript pattern: `return this.#text || (this.#text = printNode(this.#node!));`

Although, I'm not sure that it's any different performance whatsoever.  (it's more useful in `=>` fn expressions to skip the `{  }` )

Now I'm gonna spend the rest of the day wondering if the V8 JIT engine would get any difference in perf at all.

adl/core/support/codegen.ts
```diff
@@ -37,8 +37,42 @@ export function normalizeName(value: string ) {
...
-  const printer = ts.createPrinter();
-  const sourceFile = ts.createSourceFile('temp', '', ts.ScriptTarget.Latest);
-  return printer.printNode(ts.EmitHint.Unspecified, compilerNode, sourceFile);
-}
\ No newline at end of file
+/**
+ * Represents syntax for a type. Can be constructed from raw source text or from
+ * a compiler node, and converted to either lazily.
+ */
+export class TypeSyntax {
```
fearthecowboy, line 44: This is a really good idea btw. Nice.

adl/test/scenarios/v3/single/output/ablyio/aliases/Link.ts
```diff
@@ -0,0 +1,3 @@
+
+/** Links to related resources, in the format defined by [RFC 5988](https://tools.ietf.org/html/rfc5988#section-5). This will potentially include a link with relation type `next`, `first`, and `current`, where appropiate. */
+export type Link = Header<string & RegularExpression<'(<(.*)?>; rel=\"(first|current|last)?\",)*(<(.*)?>; rel=\"(first|current|last)?\")+'>, "link">;
```
fearthecowboy, line 3: This is exactly why we need to be able to have type aliases with constraints. OMFG

fearthecowboy:
I'm going to merge this right away so I can pull it in too.

PR #141, opened 2020-06-19
Use type aliases for requests definitions, make header definitions generic

nguerrera:
* Use type aliases for request bodies
* Make type aliases generic when the definition is missing some information that is only at the use.
* Introduce contrived.yaml hand-written test to stress the edge cases discussed below.

Now, when something is `$ref`ed but the target of the ref is not complete enough to be a thing in the ADL type system, we represent the target of the ref as a generic type alias and thing which ref it, instantiate it.

In this PR, there are two instances of this:

1.  In OpenAPI 3, `components/headers` define description and schema, but not the wire name, which is mapped as part of the operations response. The previous handling of this was to hack the wire name into the ref map, but this is incorrect as there could be the same header component reused with two different wire names. . 

2.  In OpenAPI 2, request bodies are represented as a parameter, but the parameter definition does not have the media type and so we make the media type a generic parameter following the same pattern as headers. I also reworked the handling of body parameters to be as other parameters, which simplified the code. In OpenAPI 3, all of the information is in a request body object, so this change simply puts the whole request body into a non-generic type alias


nguerrera:
Still have response definitions left to go.

fearthecowboy:
Merged this so I could pull it thru asap

PR #145, opened 2020-06-23
Linter and Rules

NelsonDaniel:


adl/core/model/api-model.ts
```diff
@@ -150,7 +150,7 @@ export class Files {
    * returns all the enumTypes in the API
    */
   get enumTypes() {
-    return this.files.map(each => each.getEnums().flat().map(each => new EnumType(each)));
+    return this.files.map(each => each.getEnums()).flat().map(each => new EnumType(each));
```
fearthecowboy, line 153: 👍

adl/core/linter/utils.ts
```diff
@@ -0,0 +1,376 @@
...
+/** emits c# to get the name of a property - uses nameof when it can, and uses a literal when it's an array value. */
+export function nameof(text: string): string {
+  if (text.indexOf('[') > -1) {
+    return `$"${text.replace(/\[(.*)\]/, '[{$1}]')}"`;
+  }
+  return `nameof(${text})`;
+}
+
+
+export function* getRegions(source: string, prefix: string = '#', postfix: string = '') {
```
fearthecowboy, line 339: did we need this somewhere? IIRC, this was very specific to c#

adl/core/linter/utils.ts
```diff
@@ -0,0 +1,376 @@
...
+  | Operation
+  | OperationGroup
+  | Property
+  | ParameterElement
+
+const acronyms = new Set([
+  'ip', 'os', 'ms', 'vm', //  'ssl', 'https', 'http', ''
+]);
+
+declare global {
```
fearthecowboy, line 39: we should trim out functions in this file that we're not using. some of them are not very good for ts.

adl/core/linter/utils.ts
NelsonDaniel: I missed that one. I will remove it. Thanks!

adl/core/linter/utils.ts
NelsonDaniel: Sounds good. Will do

adl/core/linter/rules/avoid-inline-types.ts
```diff
@@ -0,0 +1,36 @@
+// Need: I would like to be able to extract the definition, so it can be rehused.
```
fearthecowboy, line 1: make sure you remove the rules from this project too.

(move the test to the rules project too ?-- we should have the tests in there I think?)

fearthecowboy: although, maybe hold off a bit for moving the tests  -- it might be more work than we'd like for now.

adl/core/linter/rule.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Operation } from '../model/http/operation';
+import { OperationGroup, ParameterElement, ResponseCollection, ResponseElement, ResultElement } from '../model/operation';
+import { AliasType } from '../model/schema/alias';
+import { EnumType, EnumValueElement } from '../model/schema/enum';
+import { ModelType } from '../model/schema/model';
+import { Property } from '../model/schema/property';
+import { Declaration } from '../model/typescript/reference';
+
+export interface Rule {
+  runOn: 'edit' | 'onDemand';
```
fearthecowboy, line 11: can we rename `runOn` to `activation` -- this will help me dispatch where to send events during initialization.

and let's add a `disabled` value too -- ` activation: 'edit' | 'demand' | 'disabled' ` ;

This will make it so someone can disable their rule (or any event handler) by setting it to disabled.

fearthecowboy:
@nelsondaniel -- I'm going to merge this now, as I want to pick up a couple things.

PR #148, opened 2020-06-24
Add missing await causing duplicate type aliases

nguerrera:


adl/test/scenarios/v3/multiple/petstore/output/aliases/Pets.ts
```diff
@@ -1,3 +1,2 @@
 import { Pet } from '../models/Pet';
 export type Pets = Array<Pet>;
-export type Pets = Array<Pet>;
```
nguerrera, line 3: This would occasionally disappear like one out of 20 test runs for me due to luckier timing. It started getting on my nerves, so I stopped to fix it, took me longer than I'd like to admit to debug.

adl/core/serialization/openapi/v3/serializer.ts
```diff
@@ -68,7 +68,7 @@ export async function processOpenApi3(oai3: v3.Model, $: Context) {
     (<HttpProtocol>$.api.protocols.http).authenticationRequirements.push(security);
   }
 
-  processPaths([oai3.paths, oai3['x-ms-paths']], $);
+  await processPaths([oai3.paths, oai3['x-ms-paths']], $);
```
nguerrera, line 71: I wonder if there's a linter rule somewhere that can catch this.

fearthecowboy: OHO. I found the damn rule.

`@typescript-eslint/no-floating-promises`

I can add that in my next PR 

I was looking for that for before, but never saw it. never thought to search for 'promise' instead of `async`

nguerrera: #149

PR #154, opened 2020-06-29
Flesh out enum and model navigation

nguerrera:
The first bunch of commits are cleanup, the last two implement the change in the PR title:

* Retrieve enum values
* Retrieve model property types
* Add some test coverage of above

### DIFF
--- a/adl/core/linter/linter.ts
+++ b/adl/core/linter/linter.ts
@@ -72,7 +72,7 @@ export class Linter extends EventEmitter<Events> {
     // modelTypes and properties
     for (const modelType of files.modelTypes) {
       yield* [ ... this.iterEmit('ModelType', model, modelType)].select( each => each.result);
-      for (const property of modelType.getProperties()) {
+      for (const property of modelType.properties) {
         yield* [ ... this.iterEmit('Property', model, property)].select( each => each.result);
       }
     }
--- a/adl/core/model/schema/enum.ts
+++ b/adl/core/model/schema/enum.ts
@@ -3,14 +3,14 @@ import { EnumDeclaration, EnumMember } from 'ts-morph';
 import { literal, normalizeIdentifier, TypeSyntax } from '../../support/codegen';
 import { createDocs } from '../../support/doc-tag';
 import { ApiModel } from '../api-model';
-import { Collection, Identity } from '../types';
+import { Identity } from '../types';
 import { NamedElement } from '../typescript/named-element';
 import { SchemaInitializer } from '../typescript/schema';
 import { TypeReference } from './type';
 
 export interface EnumInitializer extends SchemaInitializer {
   extensible?: boolean;
-  readonly values: Collection<EnumValue>;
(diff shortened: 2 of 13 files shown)


adl/core/model/schema/enum.ts
```diff
@@ -87,6 +99,13 @@ export class EnumType extends NamedElement<EnumDeclaration> implements TypeRefer
     return this.node.getMembers().map(each => new EnumValueElement(each));
   }
 
+  get extensible(): boolean {
+    return hasTag(this.node, 'extensible');
+  }
+  set extensible(value: boolean) {
+    setTag(this.node, 'extensible', value);
```
fearthecowboy, line 106: do you want `@extensible` to mean `@extensible true` , (and that implies `@extensible false` is a valid statement) or should we just have the value be the presence of the `@extensible` tag?

nguerrera: The current implementation of setTag special cases booleans so it will drop the tag if value is false, and add it with no text if value is true.

fearthecowboy: Doh! riiiight!

nguerrera: I'm not sure if this is best, but this is what was left behind from the earlier manipulation that is being resurfaced here. So, `@extensible asdfasdfadsfasdf` or even `@extensible false` means extensible, the text is ignored. You can't set the text via the API. This probably needs more thought.

fearthecowboy: We can add rules in for Annotations like that so we can make errors for having the text on nodes like that.

adl/core/linter/linter.ts
```diff
@@ -72,7 +72,7 @@ export class Linter extends EventEmitter<Events> {
     // modelTypes and properties
     for (const modelType of files.modelTypes) {
       yield* [ ... this.iterEmit('ModelType', model, modelType)].select( each => each.result);
-      for (const property of modelType.getProperties()) {
+      for (const property of modelType.properties) {
```
nguerrera, line 75: (Side note on this. I'm actually a (C#/.NET design guideline biased) fan of how it was. When something creates a new array every time, it is better as a method to prevent caller from calling it over and over again, which could be expensive. However, the API has properties like this everywhere else, so I'm making it consistent here.)

fearthecowboy: certainly worth having the conversation. 

I suppose if someone used 
``` js 
for( const each in foo.getProperties ) {
  const item = foo[each];
// ....
}
```

It would be a nightmare. 

Although, nothing stops them from doing the same thing with the method either.

nguerrera: True, it doesn't stop you, but `x.getProperties()[each]`  is easy enough to spot and question whereas x.properties[each] looks more innocent. That's the argument in .NET design guidelines, anyway. I don't feel too strongly on this, and would prefer consistency above all so I went with aligning this to what's already there everywhere else.

PR #155, opened 2020-06-30
Plugin work

fearthecowboy:
- moved plugins to a subfolder, this will let us add tsconfig references for testing without accidentally making other packages potentially treated as plugins.

- plugins now build, expose their `/types` as ambient types 
- some sample stuff in the sample (still a WIP)
- renamed the extension packages to put them in an `@adl-plugins` organization


adl/plugins/core/types/constraints.ts
```diff
@@ -0,0 +1,55 @@
...
+export declare interface MultipleOf<N extends number> extends Attributes.Constraint {
+}
+
+/** a maximum value */
+export declare interface Maximum<N extends number> extends Attributes.Constraint {
+}
+
+/** a minimum value */
+export declare interface Minimum<N extends number> extends Attributes.Constraint {
+  readonly _length: N;
```
nguerrera, line 29: Just curious, why the leading underscores in these names?

fearthecowboy: good question.  The underscore was part of me experimenting. 

I wanted to put in something into the interface so that it would be able to detectable as a different type.  When we're using `string & MaxLength<5>` I want to make sure that the thing on the inside is not going to conflict with a property on which it's on. 

gonna update it.


I switched to `$length` style elsewhere, I'll make it

PR #173, opened 2020-07-15
Bug fixes

nguerrera:
1. Make sure to always pick valid identifiers

* We now use compiler API to check identifier validity, and quote and escape using JSON.stringify. 
* We no longer reject `$` unnecessarily
* We prefix reserved words with _, but allow them as TypeScript does in member names.

2. Add required imports to operation/parameter/repsonse/header declarations

* We weren't flowing the requiredReferences correctly, and we weren't calling addImportsTo
* I also batched the addition of imports as once I had them all getting emitted, we slowed down tremendously with the usual ts-morph non-batched problem.
* Overall, tests run a couple of seconds faster with all my changes now.

3. Always construct TypeSyntax from compiler nodes, never from ad-hoc strings.

* This has the side effect of making anonymous uses of model types drop doc comments. The same thing happens with anonymous enums already. It is complicated to have doc comments on anonymous types IMHO. I'd prefer if we fell back to generating names when there is documentation. I think in the long term language, we probably won't even want anonymous types. It's complex, and our use case of code generation is going to need names in the end.

* The upside is that a bug that took me a long time to debug where we had invalid content in an import is now prevented

4. Use double quotes for string literals throughout

* This is not my preference, but the compiler pretty printer was already forcing double quotes in many places, as was JSON.stringify for quoting and escaping identifiers (see 1). The path of least resistance to consistency, for now,  is to use double quotes everywhere. I think the right fix would be to have a formatter, which should arguably be user configurable handle aesthetics like that. There are other issues, such as where the pretty printer is putting newlines in the response collection types,  that could be addressed by this.


adl/core/model/api-model.ts
```diff
@@ -278,7 +278,10 @@ export class ApiModel extends Files {
       indentationText: IndentationText.TwoSpaces,
       insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces: true,
       newLineKind: NewLineKind.LineFeed,
-      quoteKind: QuoteKind.Single,
+      // It's an uphill battle to use single quotes because we are leveraging
+      // compiler and JSON API that use double quotes, so use double everywhere
+      // for consistency. At some point, we should have a separate formatting pass.
+      quoteKind: QuoteKind.Double,
```
fearthecowboy, line 284: 👍

PR #180, opened 2020-07-30
Update Tests

bterlson:
Prefer using vanilla mocha to class-n'-decorators frameworks. Also since Node 10 we've had promised fs built in, though you need the Node 14 types for it to compile properly.


bterlson:
This will look much better once #179 is merged, I didn't rebase :/

PR #181, opened 2020-08-03
Implement parser

bterlson:
This is a quick 'n dirty pass at implementing a parser for the grammar found in language.md.

There are some deltas with what was discussed last week:
* `@` is used for decorators rather than `[`/`]`. This is just an aesthetic preference.
* `.` is used for member access rather than `:`. Also an aesthetic preference.
* Implemented unions (`|`), tuples (`[...]`) and array types (`id[]`)
* The other changes should be fixes.

### DIFF
--- a/adl/language/character-codes.ts
+++ b/adl/language/character-codes.ts
@@ -186,8 +186,6 @@ export function isIdentifierPart(ch: number,): boolean {
     ch >= CharacterCodes.a && ch <= CharacterCodes.z ||
     ch >= CharacterCodes._0 && ch <= CharacterCodes._9 ||
     ch === CharacterCodes.$ || ch === CharacterCodes._ ||
-    // "-" and ":" are valid
-    ch === CharacterCodes.minus || ch === CharacterCodes.colon ||
     ch > CharacterCodes.maxAsciiCharacter && isUnicodeIdentifierPart(ch);
 }
 
--- a/adl/language/parser.ts
+++ b/adl/language/parser.ts
@@ -1,7 +1,7 @@
 import { Kind, Scanner } from './scanner';
 
 export function parse(code: string) {
-  const s = new Scanner(code);
+  const scaner = new Scanner(code);
   nextToken();
   return parseADLScript();
 
@@ -13,18 +13,18 @@ export function parse(code: string) {
       end: 0
     };
 
-    while (!s.eof) {
+    while (!scaner.eof) {
       script.statements.push(parseStatement());
     }
 
-    script.end = s.offset;
+    script.end = scaner.offset;
     return script;
   }
 
   function parseStatement(): Statement {
     let decorators = [];
 
-    // eslint-disable-next-line
+    // eslint-disable-next-line no-constant-condition
     while (true) {
       const tok = token();
       let node: Statement;
@@ -400,7 +400,7 @@ export function parse(code: string) {
     if (id !== Kind.Identifier) {
       error(`expected an identifier, got ${Kind[id]}`);
     }
-    const sv = s.value;
+    const sv = scaner.value;
 
     nextToken();
 
@@ -413,23 +413,23 @@ export function parse(code: string) {
 
   // utility functions
   function token() {
-    return s.token;
+    return scaner.token;
   }
 
   function tokenValue() {
-    return s.value;
+    return scaner.value;
   }
 
   function tokenPos() {
-    return s.offset;
+    return scaner.offset;
   }
 
   function nextToken() {
-    s.scan();
+    scaner.scan();
 
     // skip whitespace tokens for now
     while (token() === Kind.Whitespace || token() === Kind.NewLine) {
-      s.scan();
+      scaner.scan();
     }
   }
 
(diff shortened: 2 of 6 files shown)


bterlson:
Oh there is also a bug I haven't tracked down yet where a newline at the end of the file is an iloop. I suspect the scanner isn't advancing but haven't investigated yet. When testing, just make sure there is no trailing newline 😁

adl/language/character-codes.ts
```diff
@@ -186,8 +186,6 @@ export function isIdentifierPart(ch: number,): boolean {
     ch >= CharacterCodes.a && ch <= CharacterCodes.z ||
     ch >= CharacterCodes._0 && ch <= CharacterCodes._9 ||
     ch === CharacterCodes.$ || ch === CharacterCodes._ ||
-    // "-" and ":" are valid
-    ch === CharacterCodes.minus || ch === CharacterCodes.colon ||
```
bterlson, line 190: colon can't be used in an identifier if we want colons to denote fields because things like `{ x:y }` become ambiguous.

minus can't be used if we ever want to use a minus operator for anything.

I think removing these as valid identifier parts is the right approach.

bterlson:
the remaining lint error is an eslint bug :/

adl/language/character-codes.ts
fearthecowboy: Definitely colon, we don't need that in there. 

Minus may be very handy -- I've seen so very many specs that have a dash in a name.  including all of azure : `api-version` .

bterlson: For model/interface members, quoted keys work fine I think. But at the least, we can do the minimal thing now and if we want to make identifiers with dashes allowed in certain contexts later it should be doable.

bterlson: Just to expand on this for posterity - the "certain contexts" above is mostly the lhs of property declarations, because in that context there is very little allowed syntax. So something like `{ x-1: 2 }` would be doable even with a minus operator. However this property can't be referenced in expression position without either A) deciding that we won't ever have minus, B) deciding that `x-1` and `x - 1` are different (the latter being minus), or C) requiring quoted property access (which might introduce confusing inconsistencies between declaration and reference forms).

adl/language/parser.ts
```diff
@@ -0,0 +1,603 @@
+import { Kind, Scanner } from './scanner';
+
+export function parse(code: string) {
+  const s = new Scanner(code);
```
nguerrera, line 4: I'd prefer calling this `scanner`. It's used far away.

adl/language/parser.ts
```diff
@@ -0,0 +1,603 @@
...
+    }
+
+    script.end = s.offset;
+    return script;
+  }
+
+  function parseStatement(): Statement {
+    let decorators = [];
+
+    // eslint-disable-next-line
```
nguerrera, line 27:
```suggestion
    // eslint-disable-next-line  no-constant-condition
```

I think it's nice to see what eslint was reporting when we suppress.

Also, in this case, I'd be in favor of reconfiguring the rule with checkLoops=false because it's annoying that it complains on something as idiomatic as `while (true)`

adl/language/parser.ts
```diff
@@ -0,0 +1,603 @@
...
+  ): InterfaceStatementNode {
+    const pos = tokenPos();
+    parseExpected(Kind.InterfaceKeyword);
+    const id = parseIdentifier();
+    parseExpected(Kind.OpenBrace);
+    const properties: Array<InterfacePropertyNode> = [];
+
+    if (parseOptional(Kind.CloseBrace)) {
+      return <any>finishNode({
+        type: 'InterfaceStatement',
```
nguerrera, line 67: Is this supposed to be  `kind: SyntaxKind.InterfaceStatement`?

adl/language/parser.ts
```diff
@@ -0,0 +1,603 @@
...
+    pos: number,
+    node: Partial<T>
+  ): T {
+    const properties: Array<ModelPropertyNode> = [];
+
+    if (parseOptional(Kind.CloseBrace)) {
+      return <any>finishNode({
+        ...node,
+        properties
+      }, pos); // todo: 😠
```
nguerrera, line 160: Can you put more info about what is left to do in these cases?

adl/language/scanner.ts
```diff
@@ -684,7 +695,9 @@ export class Scanner {
    */
   positionFromOffset(offset: number): Position {
     let position = { line: 0, character: 0, offset: 0 };
-    if (offset < 0 || offset >this.#length) {
+
+    // eslint-disable-next-line
```
nguerrera, line 699:
```suggestion
    // eslint-disable-next-line keyword-spacing
```

(And sheesh, that's an annoying eslint bug. 🙁)

adl/language/parser.ts
bterlson: FWIW my preference is for no eslint as I don't think it provides value beyond typescript + prettier, only friction.

adl/language/parser.ts
bterlson: Great catch. Bitten by the any cast ;)

adl/language/parser.ts
bterlson: I've just fixed them. The frown made more sense before eslint made me move from suffix `as any` to prefix `<any>`.

bterlson:
Since I've changed very little since @fearthecowboy approved, I'll go ahead and merge this now.

PR #184, opened 2020-08-07
Allow semicolons as member separators

nguerrera:
Allow semicolons between interface and model members, and allow trialing commas and semicolons after interface and model members.

Also, in separate initial commit, refactor a bit to avoid having to cast to any.

### DIFF
--- a/adl/language/test/test-parser.ts
+++ b/adl/language/test/test-parser.ts
@@ -24,6 +24,11 @@ describe('syntax', () => {
        };`,
 
       `model Car {
+         prop1: number;
+         prop2: string;
+       }`,
+
+      `model Car {
           engine: V6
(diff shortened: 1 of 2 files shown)


adl/language/test/test-parser.ts
```diff
@@ -79,8 +84,10 @@ describe('syntax', () => {
 
   describe('interface statements', () => {
     parseEach([
+      'interface Store {}',
       'interface Store { read(): int32 }',
       'interface Store { read(): int32, write(v: int32): {}',
+      'interface Store { read(): int32; write(v: int32): {}',
```
nguerrera, line 90: Just spotted the missing final `}` here (before and after). Looking at it.

PR #187, opened 2020-08-07
Fix official build

nguerrera:
I'm not exactly sure what is going on here, but the official build [started failing](
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=487678&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=9c939e41-62c2-5605-5e05-fc3554afc9f5&l=388) when the test project got updated to @types/node 14. Somehow an error that only triggers when the core project use @types/node 14 started failing. When I upgraded every project to @types/node 14, I saw the error locally. 

Why is there cross-talk between project packages, but only in the official build, not in PR validation?

I'm not sure if this is an OK fix, but I've put @types/node 14 in all projects and fixed the error about a possible null ref.

### DIFF
--- a/adl/.template/package.json
+++ b/adl/.template/package.json
@@ -31,7 +31,7 @@
   "readme": "https://github.com/Azure/adl/blob/master/readme.md",
   "devDependencies": {
     "mocha": "7.1.1",
-    "@types/node": "12.7.2",
+    "@types/node": "14.0.27",
     "typescript": "~3.8.3",
     "@typescript-eslint/eslint-plugin": "2.28.0",
     "@typescript-eslint/parser": "2.28.0",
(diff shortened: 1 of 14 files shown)


bterlson:
FWIW I may eventually find a way around the node 14 req, but for now I think this is fine.

nguerrera:
In the meantime, I guess we should change these to say node 14 required?

https://github.com/Azure/adl/blob/master/documentation/Usage/readme.md
https://github.com/Azure/adl/blob/master/instructions.md

fearthecowboy:
Yeah, node 14 is fine. By the time we have this done, it will be the LTS version anyway.

PR #188, opened 2020-08-07
Added simple unit tests.

NelsonDaniel:
Added some initial work on testing rules. Right now there are some simple tests running. However, there should still be some investigation on the best way to test refactors and quick fixes. Also, to make it easier to test messages, something like ESLINT messages field could be implemented. (https://eslint.org/docs/developer-guide/working-with-rules#messageids)


nguerrera:
@fearthecowboy, should we merge this?

PR #189, opened 2020-08-07
Support alias statements

nguerrera:


adl/language/parser.ts
```diff
@@ -47,6 +47,11 @@ export function parse(code: string) {
           node = parseInterfaceStatement(decorators);
           decorators = [];
           return node;
+        case Kind.AliasKeyword:
+          if (decorators.length > 0) {
+            error('Cannot decorate an alias statement');
```
nguerrera, line 52: Following grammar here, but I think annotating aliases would be useful.

bterlson:
Should we do `type <id> = <expr>` as per the Guerrera rule (follow TypeScript syntax unless deviation has strong rationale)?

nguerrera:
Maybe. :)

@fearthecowboy I suggested yesterday a possible principle to save on debating, which @bterlson is now calling the "Guerrera" rule above. 😂 The idea is that we borrow from TypeScript wherever we don't have a significant reason to deviate.  It means inventing and debating less syntax. This would favor `@` over `[ ]` for annotations, which I know you don't like. Apart from not being different than typescript, `@` also avoids ambiguity with tuple types and  illustrates a less subjective benefit of staying close to typescript, less risk of going down a path that leads to a problem TypeScript has already avoided.

nguerrera:
This one would also be trivial to support both, but I'm not sure I want to do that for everything, could get messy. Though punting, letting people try both and tell us which they like better has its merits too.

nguerrera:
I'll merge this which matches the current grammar and can easily readjust following some discussion.

PR #191, opened 2020-08-10
Implement binder and checker

bterlson:
This is a draft PR, I've backed out many changes and am now making sure I'm starting from a base that CI likes. I had to make significant changes to everything - with this PR, we are now shipping a native ESM package!


bterlson:
ESLint actually broke code with its autofix 😥 (I had declared a type Symbol and used it many places, autofix replaced all usages with `symbol`).

adl/.default-eslintrc.yaml
```diff
@@ -24,8 +24,7 @@ rules:
     - 'error'
```
bterlson, line 24: These are just formatting changes.

adl/language/compiler/program.ts
```diff
@@ -0,0 +1,205 @@
...
+        kind: 'decorator',
+        path,
+        name,
+      });
+    }
+  }
+}
+
+
+compile('./samples/petstore').catch((e) => console.error(e));
```
bterlson, line 206: Running this file always loads the petstore sample. This is fine for now, until CLI work happens.

bterlson:
Probably good for a first review. Obviously much work is left to do, but this serves as a base - you can update petstore.adl and the decorators and play around.

adl/language/compiler/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  InterfacePropertyNode,
+  InterfaceStatementNode,
+  ModelPropertyNode,
+  Node,
+  NumericLiteralNode,
+  StringLiteralNode,
+  TupleExpressionNode
+} from './parser';
+
+export interface Type {
```
nguerrera, line 13: I like these in their own file. I think it would be good to put the syntax nodes in a file separate from the parser implementation too. (No need to do that as part of this PR, though.)

adl/language/compiler/types.ts
```diff
@@ -0,0 +1,80 @@
...
+  node: Node;
+}
+
+export interface ModelType extends Type {
+  kind: 'Model';
```
nguerrera, line 18: Should we use an enum here for consistency with the syntax nodes?

adl/language/samples/petstore/decorators.js
```diff
@@ -0,0 +1,12 @@
+// pretend I just typed this file from scratch... and saved it
```
nguerrera, line 1: nice job remembering the relative path to decorators.js :)

adl/language/samples/petstore/openapi.js
```diff
@@ -0,0 +1,310 @@
+// @ts-check
```
nguerrera, line 1: I think this is the first pure JS file of any substance I've ever reviewed. 😊

I assume it's not worth the ceremony right now to be compiling the samples from TS?

FWIW, @azure-tools/openapi would be helpful here, it has typings for openapi v2 and v3, and is what is used for importing openapi to the prior language prototype. It started in this repo and was extracted.

adl/language/compiler/parser.ts
```diff
@@ -459,7 +459,7 @@ export function parse(code: string) {
   }
 
   function error(msg: string) {
-    throw new Error(`[${scaner.position.line}, ${scaner.position.character}] ${msg}`);
+    throw new Error(`[${scaner.position.line + 1}, ${scaner.position.character + 1}] ${msg}`);
```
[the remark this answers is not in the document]
bterlson: This is just temporary, eventually `error` will create diagnostics rather than throw. The reduntant throw isn't exactly redundant, though - its to hint to TypeScript's control flow-based checker that error() results in an exception. Without it, you would need an unreachable return to satisfy the compiler.

adl/language/compiler/types.ts
bterlson: Can you say more about which types you want in which files?

My plan was to move AST node types into types.ts too (along with Kind, SyntaxKind, etc.).

adl/language/compiler/types.ts
bterlson: IMO no, because these objects are part of the object model and exposed to decorators, it needs to be more ergonomic to check what kind of node you have.

adl/language/samples/petstore/openapi.js
bterlson: Nice, thanks for the pointer!

And yeah we'll want TS eventually, but for now JS is far easier.

adl/language/compiler/types.ts
nguerrera: All in types.ts works. I could also imagine the syntax and symbol models in separate files. No strong opinion here. :)

nguerrera: Basically anything where I can review the shapes of nodes separately from implementation.

adl/language/compiler/parser.ts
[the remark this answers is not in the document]
nguerrera: >Without it, you would need an unreachable return to satisfy the compiler.

Does returning `never` not also solve that?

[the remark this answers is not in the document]
bterlson: ah, sure, you could have `return error()` instead of `throw error()` yes.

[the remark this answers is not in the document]
nguerrera: Do you need the return?

never

bterlson:
Rebased on master. Go CI!!

PR #196, opened 2020-08-17
Parse spread, make interface parameters model bodies, and implement binding/checking for templates

bterlson:
This PR introduces new syntax (spread properties for models, and interface members -> model expression bodies) and type checking for the previously introduced template types (doesn't yet support `model foo<a> =` form).

Syntax wise, we add a spread operator `...`, which is a way to combine object models together (identical to `&`, but possibly more convenient). Second, we make parameter lists model expression bodies. Combined, we can abstract common parameters using the following syntax:

```ts
model ETagHeaders {
  @description("Used to perform an operation only if the targeted resource's etag matches the value provided.")
  @header ifMatch?: string;

  @description("Used to perform an operation only if the targeted resource's etag does not match the value provided.")
  @header ifNoneMatch?: string;
}

model SyncTokenHeader {
  @description("Used to guarantee real-time consistency between requests.")
  @header syncToken?: string,
}

model ServiceParameters {
  @path endpoint: string;
  @query apiVersion: string;
}

@resource('/kv')
interface KeysResource(... ServiceParameters) {

  @description("Gets a single key-value.")
  @operationId("GetKeyValue")
  read(
    ... ETagHeaders,
    ... SyncTokenHeader,
    
    // an additional param unique to this endpoint
    @description("The key of the key-value to retrieve.")
    @path key: string,
  ): KeyValue;
}
```


bterlson:
I am loving interface parameters less and less.

adl/language/compiler/parser.ts
```diff
@@ -74,11 +74,14 @@ export function parse(code: string) {
     const pos = tokenPos();
     parseExpected(Kind.InterfaceKeyword);
     const id = parseIdentifier();
-    let parameters: Array<Types.InterfaceParameterNode> = [];
+    let parameters: Array<Types.ModelPropertyNode | Types.ModelSpreadPropertyNode> = [];
```
nguerrera, line 77: Not sure if it's worth it, but this union appears in enough places to maybe have an alias.

bterlson: If there's a third I'd say move to a union 😀

bterlson:
Somehow didn't notice the checker breakage. I'll get to this tomorrow.

bterlson:
This is moved to draft as these changes are now quantumly entangled with generics.

bterlson:
So at this point I think the branch is worth checking out and playing with. Most things in the appconfig sample work, and you can use the new built-in decorator `@inspectType` to dump the type of anything. A trick to dump the type of any type expression is to use `model =`:

```
@inspectType
model test = <whatever>
```

bterlson:
At this point I'm working on getting the decorators working again for openapi generation. I'd say these changes are good to go @nguerrera.

adl/language/lib/decorators.js
```diff
@@ -0,0 +1,14 @@
+const desc = new Map();
```
nguerrera, line 1: I'm getting an IDE warning from eslint due to this being a .js file. We should probably exclude this from config.

adl/language/compiler/parser.ts
```diff
@@ -584,3 +613,86 @@ export function parse(code: string) {
     return false;
   }
 }
+
+type NodeCb<T> =  (c: Types.Node) => T;
```
nguerrera, line 617: I am guessing "CB" is callback? Is that a pervasive acronym in these parts?

bterlson: In my experience yes but I'm happy to rename it to be more clear.

nguerrera: Up to you. I would probably have spelled out callback, but that could be my .NET bias.

bterlson:
I'll fix up that name but I'm going to merge for now.

PR #197, opened 2020-08-18
Skip comments in parser

nguerrera:
Working next on more sophistication to actually retrieve the comments and so forth, but this is a small precursor to at least unblock the use of comments by simply skipping them in the parser.

Also made dumpAST in tests convert numeric SyntaxKind to string so that it is easier to eyeball.

### DIFF
--- a/adl/language/test/test-parser.ts
+++ b/adl/language/test/test-parser.ts
@@ -1,4 +1,5 @@
 import { parse } from '../compiler/parser.js';
+import { SyntaxKind } from '../compiler/types.js';
 
 describe('syntax', () => {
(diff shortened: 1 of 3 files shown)


adl/language/test/test-parser.ts
```diff
@@ -167,7 +182,10 @@ function parseEach(cases: Array<string>) {
 }
 
 function dumpAST(astNode: any) {
-  console.log(JSON.stringify(astNode, null, 4));
+  const replacer = function(this: any, key: string, value: any) {
```
bterlson, line 185: very nice

PR #198, opened 2020-08-20
Long strings and description decorator sugar

nguerrera:
## Multi-line strings

It turns out that all ADL strings were multi-line already, we had yet to make a distinction between single quote, double quote and backtick, other than changing which would require escaping. This does not change here, but we probably want to reconsider that.

On top of that, we implement a new "triple quoted" string literal that takes inspiration from Swift "long strings" and python docstrings. At present, you can use three double quotes, three single quotes or three backticks. Again, the only difference is which you need to escape. And again, we may want to reconsider that.

```
"""
This is a long `string`

And it's easy to use **markdown** here
"""
```

Roughly specified rules:
  1. There must be a newline following opening triple quotes. This newline is not part of the string. 
  2. Whitespace between opening triple quotes and newline is ignored. 
  3. There must be a newline before the closing triple quotes. This newline is also not part of the string.
  4. The indentation of the closing triple quotes must match all other lines.
  5. The matching indentation on all lines is not part of the string
  6. Otherwise blank lines are allowed to have a subset of the matching indentation.

These are basically the same as Swift, but (2) is a relaxed exception to (1) based on my own preference. I found it unfriendly if this hard-to spot trailing whitespace before first line of content caused an error. I think we can leave it to a formatter/linter to get that picky.

We might want to consider relaxing some of the other rules too, `"""This could be valid if we want"""`. 

## Escape sequences and string values
Escape sequences were only used by the scanner to find where the literal text ends correctly, but we were not interpreting them anywhere. We trimmed off the quotes when passing to decorator, but left the rest escaped.

The scanner now provides access to the string value that unquotes, unescapes, unindents triple quoted strings. Opportunistically, I just tacked on another stringValue alongside value as there are places assuming value is the verbatim text from the source code. I would like to refactor this but I didn't want that to get mixed up with this change/proposal. 

Also opportunistically, the string value is obtained via a second pass (actually, there are multiple passes 😳) over the token text. This really shouldn't be necessary but I think it would be much easier to avoid if I did some of the scanner refactoring I have in mind first.

We only support `\r \n \t \' \" \' and \\` with this change. I would lean slightly towards adding only `\u{<code point>}` to cover the rest, but I didn't implement that here and I think filling this out is best left as its own issue to design and discuss.

## Description decorator sugar
We allow an unadorned `"string literal"` as sugar for `@description("string literal")`

This can appear before explicit decorators or after members, currently. Annoyingly, the leading version must be triple-quoted for model properties to avoid ambiguity vs. quoted property name.

For now, I haven't allowed trailing explicit decorators, only trailing short descriptions.

```
"""
Vehicle to get from point A to B.

See also `Bicycle`
"""
model Car {
   engine : Engine        "Makes the car go.";
   color: Color           "The color of the car's exterior paint.";

   """
   You can also use long descriptions before properties
   """
   property: Type;
}
```

## Scanner error handling
I replaced scanner.state (that we never checked) with scanner.onError handler that the parser currently wires to it's current error handler that just throws for now. I did still try somewhat to keep recovery in mind following errors I added to the scanner, but these paths are not exercised currently and will need lots more work.


nguerrera:
@bterlson Calling this ready for review. Played with it a bit manually vs. some appconfig changes, added a few more tests, and fixed up some issues. Still would like to add more tests and spec it better, but I think I'll follow up with those separately.

adl/language/test/test-scanner.ts
```diff
@@ -120,8 +121,46 @@ describe('scanner', () => {
...
+    scanString(
+      `"""   
+      This is a triple-quoted string
+
+  
+      
+      And this is another line
+      """`,
+      // NOTE: sloppy blank line formatting and trailing whitespace after open
+      //       quotes above is deliberately tolerated.
```
nguerrera, line 157: Feeling even better about this call seeing that the slop is invisible in GitHub reviews.

.vscode/launch.json
```diff
@@ -11,9 +11,23 @@
       "port": 9229,
       "smartStep": true,
       "sourceMaps": true,
+      "skipFiles": [
```
bterlson, line 14: ha I have this exact change in my worktree!

adl/language/compiler/checker.ts
```diff
@@ -302,7 +302,7 @@ export function createChecker(program: Program) {
     const value =
       node.kind === SyntaxKind.NumericLiteral
         ? Number(node.value)
-        : node.value.slice(1, -1);
+        : node.value;
```
bterlson, line 305: This will likely cause issues downstream (string values will have their quotes). This is probably fine for now.

adl/language/compiler/parser.ts
```diff
@@ -1,9 +1,11 @@
+import { format } from './messages.js';
 import { Kind, Scanner } from './scanner.js';
 import * as Types from './types.js';
 
 
 export function parse(code: string) {
-  const scaner = new Scanner(code);
```
bterlson, line 6: wow lol

nguerrera: I kind of like saying scayner now.

adl/language/compiler/checker.ts
nguerrera: I may be missing something here. The value in the StringLiteralNode should be unquoted and unescaped already with my changes.

PR #199, opened 2020-08-20
OpenAPI emit support

bterlson:
With these changes, you can emit OpenAPI. If you want to play with this:

1. Check out and buildhttps://github.com/Azure/adl/pull/199/files
2. Run dist/compiler/program.js <sample folder name> (or hit F5 in vscode)

And if everything went well, you should have a bunch of OpenAPI spew.

### Remarks

This PR introduces a couple big changes. First, caching is back in the checker during template instantiation. Having duplicate type instances for the same type makes doing anything with types super confusing (e.g. simple "have I seen this type before" calculations have to be structural comparisons). Now, every (AST node, template instantiation) tuple has a unique type.

Second, when decorators are applied to templates, they are executed on instantiations of that type, and not the uninstantiated type. This makes dealing with metadata more convenient in decorators - you just ask the type itself for its metadata. And when decorators ask about the fields they're attached to, they see instantiated types not template type placeholders. So, consider the following  ADL:

```
@inspectType
model Box<T> {
  contents: T
}

model Shelf {
  item1: Box<int32>;
  item2: Box<string>;
  item3: Box<string>;
}
```

The inspectType decorator is called twice, once for each unique instantiation, and the contents property is either `int32` or `string`. Getting these semantics required moving decorator execution into the checker. This feels wrong but has upshots - e.g. decorator exceptions effectively become type errors, it's trivial to add decorators to any type, and decorators can modify the type they're called with without any special handling from us.

Lastly, this PR completely rewrites the openapi emitter. The previous implementation was horrifying. The new version is smaller, faster, and more capable. Notably, thanks to the above changes in the checker, references are emitted far more reliably (and with an obvious place to adjust whether references or inline schemas are used).

### To do

* [ ] Checker: use instantiation scope stack with node caches, eliminate usage of MultiMap
* [ ] Checker: simplify lookup of instantiated template params by setting template parameter bindings in the template's symbol table before instantiation
* [ ] OpenAPI emit: support string literal unions or some enum pattern (app config)
* [ ] OpenAPI emit: support x-ms- extensions
* [ ] OpenAPI emit: @​lro decorator
* [ ] OpenAPI emit: investigate paging (e.g. @​paginated)
* [ ] OpenAPI emit: default operationId by to the property type's name for non-standard operation names (e.g. `sendStatus()`).
* [ ] OpenAPI emit: support for parameter arrays (e.g. `@query filterFields: string[]`).
* [ ] protobuf emit: basic support
* [ ] Language/protobuf emit: support for default values
* [ ] Language/protobuf emit: enum declarations
* [ ] Tests & cleanup


adl/language/compiler/checker.ts
```diff
@@ -1,43 +1,83 @@
...
   UnionExpressionNode, UnionType
 } from './types.js';
 
+/**
+ * A map keyed by a set of objects. Used as a type cache where the base type
+ * and any types in the instantiation set are used as keys.
+ *
+ * This is likely non-optimal.
+ */
+export class MultiMap<T> {
```
nguerrera, line 27: I'm more accustomed to `MultiMap` meaning "each key can have multiple values", than each value has a composite key. I think that's pretty standard too: https://en.wikipedia.org/wiki/Multimap

Maybe `MultiKeyedMap` or something here.

bterlson: Oh interesting. I am now retroactively embarrassed for my many misuses of this term over the years.

PR #201, opened 2020-08-25
Check bools

nguerrera:


adl/language/lib/openapi.ts
```diff
@@ -423,6 +423,7 @@ function createOAPIEmitter(program: Program) {
     switch (adlType.kind) {
       case 'Number':
       case 'String':
+      case 'Boolean':
         return adlType.value;
```
nguerrera, line 427: This needs more work. We end up with invalid swagger for properties of literal type. However, same issue pre-exists for number and string. I started fixing it and decided to pause and put this PR up, which I think is strictly forward progress as I need to context switch onto something else for most of today.

bterlson:
I like it. Thanks for the placeholder for scratch as well.

adl/language/lib/openapi.ts
bterlson: I just found out from Yanni how to wncode constants.  I can pick this up today I think.

nguerrera: Is it enum with one value?

Another issue I found is we're trying to use oneOf for unions, which doesn't appear to be valid in OAI2. I didn't dig too much, but it looks like we're limited to enum where everything can have the same type.

bterlson: Enum with one value yeah.

oneOf is indeed not supported in swagger/OAPI2 and could probably be removed from the output for now. allOf IS supported, though, and it is interesting to consider whether we should be emitting `model A { ... B, c: string }` as the flattened version like today or with an allOf B.

PR #204, opened 2020-08-31
Hack until OpenAPI 2 is valid for petstore and appconfig

nguerrera:
I mentioned last week in passing that I had seen some issues in the OpenAPI 2 emit, but didn't have a list, and it wasn't clear if all of them were known or not.

I set out to put together the list here. I've also fixed them all in a hurry in this PR as that ended up being the best way to enumerate them. However, some of the fixes are opportunistic and inefficient, and should probably be done better.

My methodology here was:

* generate petstore and appconfig
* paste output into https://editor.swagger.io/
* fix issues it flags
* repeat until all issues are gone

And these are all of the issues that I found this way:

1. Typo: s/requred/required/
2. title and version are required
3. response description is required
4. path parameter not emitted
5. some parameters duplicated
6. non-body parameters have to inline their schema, don't have schema property, can't ref schema. Also can only be of certain types, but I didn't validate that here.
7. Typo: s/Array/array
8. `required` array must have at least one element (or not be defined)
9. constants should be enum of one element
10. oneOf is not valid in OpenAPI v2 (fix: use enum for literals of same type, error otherwise.)
11. Typo: #/parameters instead of #/definitions to ref schemas
12. names of parameters and schemas with symbols like `[`  and `<` caused invalid $refs that would have to be URL encoded. (URL encoding was ugly and for some reason caused the swagger editor to still have resolution problems, so I converted the names with some hacky search and replace and counter bumping instead.)
13. parameters with same names but different defintions (type or location) were getting shared
14. some primitives returned wrong type

You can see the output delta before this PR and after it here: 

https://github.com/nguerrera/adl-oai2-diff/commit/8a7d19b21c88d1eae616f7faf9a9ae42c912f144

Side notes:
* I wanted to use `@azure-tools/openapi` to get typing to help me fix some of these, but I gave up fighting to get the module to resolve after a little while. We should push on this, though. I also found a couple of bugs in @azure-tools/openapi and I will send a PR to fix those.
* It would also be good to have an automated test that runs a swagger validator.

### DIFF
--- a/adl/language/compiler/checker.ts
+++ b/adl/language/compiler/checker.ts
@@ -130,6 +130,10 @@ export function createChecker(program: Program) {
         return (<UnionType>type).options.map(getTypeName).join(' | ');
       case 'Array':
         return getTypeName((<ArrayType>type).elementType) + '[]';
+      case 'String':
+      case 'Number':
+      case 'Boolean':
+        return type.value.toString();
     }
 
     return '(unnamed type)';
--- a/adl/language/lib/openapi.ts
+++ b/adl/language/lib/openapi.ts
@@ -12,16 +12,6 @@ import {
 } from './rest.js';
 
 
-let init = false;
-
-const root = {
-  swagger: '2.0',
-  info: {},
-  schemes: ['https'],
-  paths: {},
-  definitions: {},
-  parameters: {},
-};
 export function onBuild(p: Program) {
   const emitter = createOAPIEmitter(p);
   emitter.emitOpenAPI();
@@ -35,7 +25,10 @@ export function operationId(program: Program, entity: Type, opId: string) {
 function createOAPIEmitter(program: Program) {
   const root: any = {
     swagger: '2.0',
-    info: {},
+    info: {
+      title: '(title)',
+      version: '0000-00-00'
+    },
     schemes: ['https'],
     paths: {},
     definitions: {},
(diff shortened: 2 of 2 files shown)


bterlson:
Looks good so far! I was thinking about using a regexp to convert to a swagger-safe name as well. I like the compiler's formatting right now (getTypeName would get used for e.g. quickinfo on hover).

nguerrera:
cc @fearthecowboy

I think there are some good API we can reuse from autorest's repertoire to generate safe names. I was unsuccessful at pulling in modules from `@azure-tools` yesterday. Something to do with my complete lack of understanding at this point in my TS journey of the different module systems in JavaScript. 😳 I would like to start being able to do that for stuff like that and to type the OpenAPI we emit.

I also wonder if any strategy for naming these instantiated types is going to end up generating good code. I have a feeling folks are going to end up having to name a lot of things themselves in order to clean up their output. If this is true, and I don't know if it is, I start to wonder if there are simplifications to the language that might be possible. For example, what if template instantiations could only occur on the right-hand-side of `model Name = ...`? Templates become like macros in this world and the type system can actually be nominal. I think. Maybe this isn't a good idea,  it's just a passing thought I had, but it illustrates the tension I was feeling while trying to fix those names:

* If I have to choose the name, have I already lost?
* If the user has to choose the name in practice, does it make sense to maintain a type system where you can have complex, unnamed instantiated types?

nguerrera:
Another approach more targeted at the specific problem of naming these instantiated types in swagger is to keep them inlined, and only use #/definitions and #/parameters for user-named things. @fearthecowboy suggested this on a call today.

nguerrera:
Took this out of draft after talking to @bterlson. There's some things that I would like to clean up but this is a step forward. Feel free to approve and merge.

PR #206, opened 2020-09-24
Emit better names to OpenAPI

nguerrera:
See https://github.com/nguerrera/adl-oai2-diff/commit/df2e52240028ad01b3927027fce11c0e17443649 for the appconfig/petstore emit diff. I haven't tried autorest yet, but to my naked eye the swagger is looking a lot closer to something I might have written by hand.

1. Model `A & B` and `{ ...A, ...B }` as having "base models" A and B. These are now semantic sugar for the same type.

2. When emitting schema, if possible, reduce models down to single base model that has actual schema properties where possible. For example, in `OK<T>`, once you put the statusCode elsewhere you're left with T so use that.

3. Emit "base models" using OpenAPI allOf.

4. Instead of name munging, any type with an unspeakable name in OpenAPI simply gets inlined. Users can make type aliases with good names to avoid this inlining if desired.

5. Inline parameters that are inlined in ADL. To end up in `#/parameters`, a parameter has to be shared via spreading in ADL. This allows us to pick a good name for the `#/paramters` map: if the spread type has only one property, use the type name, otherwise use `(type name).(property name)`. It also gives the OpenAPI output a structure that mirrors the ADL.

6. Fix issues with responses. We were not emitting status codes other than 200, and not emitting response headers.

### DIFF
--- a/adl/language/compiler/checker.ts
+++ b/adl/language/compiler/checker.ts
@@ -146,9 +146,10 @@ export function createChecker(program: Program) {
       // template instantiation
       const args = model.templateArguments.map(getTypeName);
       return `${model.name}<${args.join(', ')}>`;
-    } else if (model.intersectionMembers) {
-      // intersected model
-      return model.name || model.intersectionMembers.map(getTypeName).join(' & ');
+    } else if (model.ownProperties.size == 0 && model.baseModels.length > 0) {
+      // intersection or spread-only
+      // (NOTE: might have been spelled `{ ...A, ...B }` in source but we use the terser syntax here.)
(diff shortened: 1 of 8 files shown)


nguerrera:
Looking at the output. There appears to be a bug in the recursion that simplifies down to type with schema properties. I’ll look into it in the morning.

nguerrera:
Working on bugs, converted to draft.

nguerrera:
Should be fixed up now. I updated the link to emit diff and the description to include an additional bug fix around responses that was not related to naming.

fearthecowboy (approved):

PR #211, opened 2020-11-17
Alternative parsing

fearthecowboy:
A different parsing model that builds live objects that can be round tripped.

(see /implementation.md)


daviwil:
Closing for now!

PR #223, opened 2021-01-05
Added gist as document in repo

johanste:
Source gist: https://gist.github.com/bterlson/8f36c62ee1aed27d676020762b24d20c

### DIFF
--- a/documentation/Specifications/arm.md
+++ b/documentation/Specifications/arm.md
@@ -1,0 +1,366 @@
+# Building ARM endpoints in ADL
+
+ADL provides a number of useful building blocks to make defining your ARM API endpoints super easy and consistent. This document will get you started on understanding the basics of ADL and how to define a new ARM API for your service.
+
+Summary of required primitives:
+
+* ADL Built-ins
+  * Page\<T\> and @pageable
+  * Customizable visibility modifiers
+* REST Bindings
+  * Decorators for each REST verb (@​get, @​set, etc.)
+  * Decorators for each request and response part: @body, @header, @status.
+  * Response types for different status codes (Ok<T>, Created<T>, etc.)
+* Azure Bindings
+  * LRO patterns for RELO and Stepwise
+* ARM bindings
+  * Models for common ARM resource types and request and response shapes
+  * Models for common data types within arm:
+    - secret
+    - certificate
+  * Decorators for common ARM API shapes:
+    - Resources, child resources, proxy resources, and extension resources
+    - subscription and tenant resources
+    - Custom actions
+    - Private link
+ 
+ 
+## ADL built-ins
+
+These are provided by ADL itself and are used by many ARM endpoints.
+
+### @pageable & Page\<T\>
+
+This decorator indicates to code generators that the endpoint is server-side pageable.
+
+```
+model Page<T> {
+  value: T[];
+  nextLink: string;
+}
+
+interface Books {
+  @pageable list(): Page<{ title: string }>;
+}
+```
+
+Can customize with the @pageable decorator.
+
+```
+interface BookPage {
+  text: Book[];
+  nextPage: string;
+}
+
+interface CustomBooksByPage {
+  @pageable({values: 'text', nextLink: 'nextPage'}) list(): BookPage;
+}
+```
+
+Supporting client-side pageable in a similar way should be possible.
+
+
+### Custom visibility modifiers
+
+Visibility is meant to be a user-defined constraint on properties that are visible when the model is used in particular contexts (e.g. as a request body for a PUT method). Lack of a visibility constraint implies no constraint (that is, a property appear in all contexts). Visibility may have arbitrary string values - different packages of ADL tools may want to define particular values for visibility and their definitions. ADL provides the following decorators:
+
+- @visibility, which takes a string, and flags a field as only being visible if that visibility flag is set
+- @withVisibility, which takes a list of visibility flags.
+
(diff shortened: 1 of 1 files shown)


documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+    - Private link
+ 
+ 
+## ADL built-ins
+
+These are provided by ADL itself and are used by many ARM endpoints.
+
+### @pageable & Page\<T\>
+
+This decorator indicates to code generators that the endpoint is server-side pageable.
```
johanste, line 34: the `@pageable` decorator is not expressive enough in this example (I assume it is incomplete). There are several different strategies that are employed, including continuation-token-in-header, calling a separate method with a body for the next page, relative paths vs. query parameter etc. that require more information in order to complete. In other words, more work is needed. Incidentally, we just added additional customer paging strategy support to autorest, so there should be lessons to be taken from there with regards to which scenarios to support.

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+
+```
+model Page<T> {
```
johanste, line 37: I would like to see if we could omit the paged model definitions. The current examples require service teams to know how to correctly model a paged operation response (e.g. provide an absolute `nextLink` attribute in the response). Ideally, in the vast majority of cases (at least where there is no additional per-page data), the model can be inferred from the fact that the operation is paged, possibly combined with how you parameterized the `paged` decorator. Additionally, responses can be validated against the response to ensure that they are conforming to the implicit schema (although it is hard to determine if the absence of a nextLink means that the service only has a single page of responses or if the service simply is broken). The example also doesn’t cover how the next link should be formatted (e.g. $skip/$top vs. $skiptoken etc.).

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+  name: string;
+}
+
+@WithVisibility('write')
+model WriteCar {
+  ... Car;
+}
+```
+
+For Arm, the visibility flags are:
```
johanste, line 84: ARM (and many other services) also have properties that are required for a PUT to succeed if the PUT is an update, but optional otherwise. 

The visibility modifier also doesn't define *how* patch works - both which content type is used as well as custom behaviors. For example, in ARM, a PATCH of the `tags` attribute will replace all tags, whereas a PATCH on the `identity` attribute will work like json-merge-patch. Part of this I would assume is related to REST/HTTP mapping of the description, but the content-type (which lives in the HTTP domain of things) is intertwined with the model that describes the change request and, consequently, service capabilities. I may be able to insert items into an array (which implies json-patch rather than merge patch at the HTTP layer), but not remove the whole array (as opposed to removing the items in the array)

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+ create = property can be set in PUT requests creating the undelrlying resource
+ update = property can be set in PUT requests updating the underlying resource
+ patch = property can be set in PATCH requests updating the underlying resource
+```
+
+## REST bindings
+
+These bindings imbue ADL with the ability to describe REST API endpoints. When defining ARM APIs, you won't need to use these much, but they will be useful for custom actions as you'll see later.
+
+* Decorators for each REST verb (@​get, @​set, etc.)
```
johanste, line 97: `@set`? I assume this is a misspelled `@put` :)

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+```
+model VirtualMachine {
+  id: string;
+  state: 'creating' | 'created' | 'errored';
+}
+
+@path('/virtualMachines')
+interface VirtualMachines {
+  // put semantics
+  @relo @put createVM(@path id: string): {}; // no return info
```
johanste, line 117: I'm not quite sure I follow this example. I would not expect there to be two (service) operations. Nor would I expect there to be two client library methods. What is the purpose of the RELO method with no return value?

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+@WithVisibility('write')
+model WriteCar {
+  ... Car;
+}
+```
+
+For Arm, the visibility flags are:
+
+```
+ read = property occurs in responses containing the model
```
johanste, line 87: I would suggest that we look at how to best decorate key properties (which are inherently not mutable after a resource has been created) and if/how they would interact with the visibility concept.

documentation/Specifications/arm.md
```diff
@@ -0,0 +1,366 @@
...
+    - subscription and tenant resources
+    - Custom actions
+    - Private link
+ 
+ 
+## ADL built-ins
+
+These are provided by ADL itself and are used by many ARM endpoints.
+
+### @pageable & Page\<T\>
```
johanste, line 32: After chatting with @bterlson, I'd like to propose an alternative to `@pageable`:

A `@list` decorator on an operation means that it is a semantic list operation:

```
interface Books {
    @list list(): Book 
}
```

Ideally, I wouldn't have to explicitly specify the exact shape of the response here - I only care about the fact that I'm listing `Book`s. 

List operations are implicitly assumed to be (server driven) pageable, and you'd have to explicitly opt-out to not support server-driven paging (but I'd honestly be fine with not allowing that for `@list` operations)

```
interface Books {
    @list listBooks(pagingStrategy="none"): Book 
}
```

PR #224, opened 2021-01-05
Reorganize repo to focus on a single implementation

### LINKED ISSUE #221: Archive/remove legacy prototype code and folders
There is a lot of conflicting code and documentation in this repository that needs to be cleaned up.

daviwil:
This repo contains two different prototypes of the ADL language.  Since we've picked a direction to move forward with, it made sense to move some of the code to an `.archive/` folder in the repo for the time being until we decide we no longer need to draw ideas from it.

The majority of the changes here are file moves but there are some minor changes and additions here and there (mostly toward the end of the list of changed files) to update `rush.json`, project references, and CI pipelines.

It might be helpful to browse the repo for the branch I've been working on to see what the new layout looks like:

https://github.com/Azure/adl/tree/reorganize

Fixes #221.


bterlson:
I like it. But should `./adl` be `./src` maybe?

I also wonder if we should prepare for monorepo with lerna or rush and have a top-level packages folder? If we want to use npm to distribute adl extensions/bindings/whatever, we may want to have a blessed set in the repo. E.g. a few I can think of:

* azure/adl - the core language
* azure/adl-rest - official rest API bindings (@​get/@​post/etc, @​header, @​path, @​resource)
* azure/adl-grpc - official grpc bindings
* azure/adl-gql - official graphql bindings
* azure/adl-autorest - bindings and codegen for existing autorest toolchain (supporting the ADL→swagger→autorest toolchain).
* azure/adl-rpaas - RPaaS bindings and codegen

nguerrera:
Looks good. Pulled this down and played with it a bit. I found that the mocha test explorer was giving an error and seemed to be due to a stale pointer to adl/core project. This fixed it: https://github.com/Azure/adl/compare/reorganize...nguerrera:test-explorer-fix

nguerrera:
> But should ./adl be ./src maybe?

+1, adl/adl has always thrown me for a loop. :)

daviwil:
@bterlson Yep, I sent you an IM about that earlier today :)  Changing the monorepo structure was my next task, just wanted to get this cleanup out of the way first.

Are you suggesting the top-level packages folder start with `azure/` or `packages/azure/...` ?  Personally I think `packages/` is better then we can have further specialization inside of it depending on what all we need,

@nguerrera Thanks!  I'll integrate that fix.

nguerrera (approved):

bterlson:
Sorry that was unclear, those were package names not file paths. Ignore the `azure` prefix for a strawman for what goes under `packages`, but I'm not married to any of it.

bterlson (approved):

daviwil:
OK! I'll pick up Nick's fix and then do the final moves of everything into `packages/`.  Thanks!

Issue #216, opened 2020-12-17
Add an initial command line interface for the ADL compiler


daviwil:
The ADL compiler needs a command line interface that enables a user to compile `.adl` documents into Swagger files and server-side stubs.

- [x] `adl compile adl-folder/` 
- [x] Specify output path for server side stubs

daviwil:
Fixed by #228.

PR #228, opened 2021-01-12
Add basic ADL command line interface

daviwil:
This PR adds a basic CLI interface for the ADL compiler using the `yargs` library.  Here's an example of invoking it:

```
adl compile samples/ --output-file=swagger.json
```

I've introduced a new interface for `CompilerOptions`, let me know if you think there's a better way to express this!

Part of #216.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -2,12 +2,14 @@ dependencies:
   '@rush-temp/adl.language': 'file:projects/adl.language.tgz'
   '@types/mocha': 7.0.2
   '@types/node': 14.0.27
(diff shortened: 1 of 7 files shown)


bterlson:
I think this is great for now! Very handy.

Maybe this isn't news, but openapi.ts should hopefully be a regular old plugin/extension at some point, and so would need a way to "register" output types (and maybe languages or other facets thereof), and the CLI would need a way to enable/disable them. Just something to keep in the back of your head 😀

nguerrera (approved):

daviwil:
Sounds good!  I'll start trying to spec out some of that soon.

PR #238, opened 2021-01-13
Sample adl and swagger for Liftr.Confluent service

markcowl:
This is the sample presented at the MarkRuss demo.  Let me know if there is a more appropriate placement in the repo.


daviwil (approved):
Looks good, thanks!

PR #241, opened 2021-01-13
Add decorators for HTTP verbs (get, put, post, patch, delete)

### LINKED ISSUE #231: Add HTTP action decorators to REST library
We need decorators for the standard CRUD operations that get used for resource types.  An example:

- [ ] `@rest`
- [ ] `@get`
- [ ] `@put`
- [ ] `@post`
- [ ] `@delete`
- [ ] `@patch`
- [ ] `@path` - Operation parameters, specifies that the param is part of the path

```
@rest('/subscriptions/{subscription}/resourceGroups/{resourceGroup}/providers/Microsoft.Confluent/organizations/{name}')
interface Organization 
{
  @get get(@path subscription: string, @path resourceGroup: string, @path name: string) : ArmResponse<Organization>;

  @put createOrUpdate(@path subscription: string, @path resourceGroup: string, @path name: string, Organization) : ArmResponse<Organization>;
  @patch update(@path subscription: string, @path resourceGroup: string, @path name: string, OrganizationUpdate) : ArmResponse<Organization>;
  
  @delete delete(@path subscription: string, @path resourceGroup: string, @path name: string) : ArmResponse;

  @post getKeys(@path subscription: string, @path resourceGroup: string, @path name: string) : ArmResponse<OrganizationKeys>;
}
```

daviwil:
This change adds decorators for a few HTTP verbs:

- `@get`
- `@put`
- `@post`
- `@patch`
- `@delete`

I've kept the existing verb inferrence if the author doesn't add one of these decorators to an interface method.  Also not adding tests just yet, I'm having trouble getting Jest set up on this project (considering just doing my own thing with Mocha for now...)

Resolves #231.

### DIFF
--- a/packages/adl/lib/rest.ts
+++ b/packages/adl/lib/rest.ts
@@ -65,3 +65,43 @@ export function isBody(entity: Type) {
   return bodyFields.has(entity);
 }
 
+export type HttpVerb = "get" | "put" | "post" | "patch" | "delete";
+
+const operationVerbs = new Map<Type, HttpVerb>();
+
+function setOperationVerb(entity: Type, verb: HttpVerb) {
+  if (entity.kind === "InterfaceProperty") {
+    if (!operationVerbs.has(entity)) {
+      operationVerbs.set(entity, verb);
+    } else {
+      throw new Error(`HTTP verb already applied to ${entity.name}`);
+    }
+  } else {
+    throw new Error(`Cannot use @${verb} on a ${entity.kind}`);
+  }
+}
+
+export function getOperationVerb(entity: Type): HttpVerb | undefined {
+  return operationVerbs.get(entity);
+}
+
+export function get(program: Program, entity: Type) {
+  setOperationVerb(entity, "get");
+}
+
+export function put(program: Program, entity: Type) {
+  setOperationVerb(entity, "put");
+}
+
+export function post(program: Program, entity: Type) {
+  setOperationVerb(entity, "post");
+}
+
+export function patch(program: Program, entity: Type) {
+  setOperationVerb(entity, "patch");
+}
+
+// TODO: How do we deal with reserved words?
+export function _delete(program: Program, entity: Type) {
+  setOperationVerb(entity, "delete");
+}
--- a/packages/adl/lib/rest.ts
+++ b/packages/adl/lib/rest.ts
@@ -101,7 +101,7 @@ export function patch(program: Program, entity: Type) {
   setOperationVerb(entity, "patch");
 }
 
-// TODO: How do we deal with reserved words?
+// BUG #243: How do we deal with reserved words?
 export function _delete(program: Program, entity: Type) {
   setOperationVerb(entity, "delete");
 }
--- a/packages/adl/lib/openapi.ts
+++ b/packages/adl/lib/openapi.ts
@@ -9,7 +9,9 @@ import {
   getPathParamName,
   getQueryParamName,
   getResources,
-  isBody
+  isBody,
+  getOperationVerb,
+  HttpVerb
 } from './rest.js';
 
 export function onBuild(p: Program) {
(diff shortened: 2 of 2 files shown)


daviwil:
An example:

```js
@resource("/resources/organization")
interface Organization {
  @get getKeys(): Ok<string>;
  @post postKey(): Ok<string>;
  @put updateKey(): Ok<string>;
  @_delete deleteKey(): Ok<string>;
}
```

Output:

```json
{
    "swagger": "2.0",
    "info": {
        "title": "(title)",
        "version": "0000-00-00"
    },
    "schemes": [
        "https"
    ],
    "paths": {
        "/resources/organization": {
            "get": {
                "consumes": [],
                "produces": [
                    "application/json"
                ],
                "parameters": [],
                "responses": {
                    "200": {
                        "schema": {
                            "type": "object",
                            "properties": {},
                            "x-adl-name": "{}"
                        },
                        "description": "Success"
                    }
                }
            },
            "post": {
                "consumes": [],
                "produces": [
                    "application/json"
                ],
                "parameters": [],
                "responses": {
                    "200": {
                        "schema": {
                            "type": "object",
                            "properties": {},
                            "x-adl-name": "{}"
                        },
                        "description": "Success"
                    }
                }
            },
            "put": {
                "consumes": [],
                "produces": [
                    "application/json"
                ],
                "parameters": [],
                "responses": {
                    "200": {
                        "schema": {
                            "type": "object",
                            "properties": {},
                            "x-adl-name": "{}"
                        },
                        "description": "Success"
                    }
                }
            },
            "delete": {
                "consumes": [],
                "produces": [
                    "application/json"
                ],
                "parameters": [],
                "responses": {
                    "200": {
                        "schema": {
                            "type": "object",
                            "properties": {},
                            "x-adl-name": "{}"
                        },
                        "description": "Success"
                    }
                }
            }
        }
    },
    "definitions": {},
    "parameters": {}
}
```

packages/adl/lib/rest.ts
```diff
@@ -65,3 +65,43 @@ export function isBody(entity: Type) {
...
+export function post(program: Program, entity: Type) {
+  setOperationVerb(entity, "post");
+}
+
+export function patch(program: Program, entity: Type) {
+  setOperationVerb(entity, "patch");
+}
+
+// TODO: How do we deal with reserved words?
+export function _delete(program: Program, entity: Type) {
```
daviwil, line 105: Can't name this function `delete` because it's a reserved word.  Is there any syntax we can use to expose these functions to ADL without any naming tweaks?

nguerrera:
Unrelated, but looks like there's an existing bug on the schema of the responses. I think it's my fault. We seem to be dropping string down to object. I did some tricks to try to simplify types and I think it's not stopping at the intrinsic types and seeing it as just another `{}`. It's been a while so I may not be remembering right, but I can take a look.

packages/adl/lib/rest.ts
nguerrera: Good catch. I don't think we have anything for this. @bterlson, do we?

nguerrera (approved):

daviwil:
I've already got a fix for it in what I'm working on for the next PR, let me see if what I send out next makes sense!

nguerrera:
Oh great, I just filed #242 for it. Now assigned to you.

packages/adl/lib/rest.ts
daviwil: Filed issue #243 to track this and referenced the bug number in a comment above the line.

daviwil:
Thanks! I'll go ahead and merge this and then send out the next PR this morning.

packages/adl/lib/rest.ts
bterlson: Oh interesting! Yeah, we don't, but we should probably. We could say that leading underscores are a convention that means "project this into ADL without the leading underscore"?

Issue #242, opened 2021-01-14
Dropping primitive/intrinsic response to object in opeanapi emit


nguerrera:
See https://github.com/Azure/adl/pull/241#issuecomment-759844791

daviwil:
Fixed by #244.

Issue #240, opened 2021-01-13
Add @format decorator


daviwil:
This decorator allows specifying a regular expression that is used for validating a string-based model type.

daviwil:
Fixed in #244.

PR #244, opened 2021-01-15
Improve templated intrinsic types in OpenAPI emitter, add more decorators

### LINKED ISSUE #229: Add minLength and maxLength decorators
This requires decorators and metadata for specifying length requirements on string types.

daviwil:
This change addresses #242 and resolves #229 and #240:

- Templated types like `Ok<string>` or `Ok<someTypeAssignedFromIntrinsic>` will now be correctly represented in operation response schemas
- Added `@format` decorator for pattern-based validation of string parameters
- Added `@minLength` and `@maxLength` decorators for validation of string parameter length

**Questions**

- Should `@format` be `@pattern` or something else?  I was following the naming convention in @markcowl's `confluence.adl` example

**Contrived Example**

**`sample.adl`**:

```
@doc "Success"
model Ok<T> {
  @header statusCode: 200;
  ... T;
}

@doc "Client Error"
model ClientError<T> {
  @header statusCode: 400;
  ... T;
}

@resource("/resources/organization")
interface Organization {
  @get getKeys(): Ok<email>;
  @post postKey(addr: email): Ok<string>;
  @put updateKey(key: string25): ClientError<int64>;
  @_delete deleteKey(key: string50): Ok<string>;
}

@doc("Reusable representation of an email address")
@format("\\w+@\\w+\\.\\w+")
model email = string;

@doc('Shorthand for setting length limit.')
@minLength(5)
@maxLength(50)
model string50 = string;

@doc("Shorthand for setting length limit.")
@maxLength(25)
model string25 = string;
```

**`openapi.json`**:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "(title)",
    "version": "0000-00-00"
  },
  "schemes": [
    "https"
  ],
  "paths": {
    "/resources/organization": {
      "get": {
        "consumes": [],
        "produces": [
          "application/json"
        ],
        "parameters": [],
        "responses": {
          "200": {
            "schema": {
              "type": "string",
              "pattern": "\\w+@\\w+\\.\\w+"
            },
            "description": "Success"
          }
        }
      },
      "post": {
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "addr",
            "in": "body",
            "required": true,
            "schema": {
              "type": "string",
              "pattern": "\\w+@\\w+\\.\\w+"
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "type": "string"
            },
            "description": "Success"
          }
        }
      },
      "put": {
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "key",
            "in": "body",
            "required": true,
            "schema": {
              "type": "string",
              "maxLength": 25
            }
          }
        ],
        "responses": {
          "400": {
            "schema": {
              "type": "integer",
              "format": "int64"
            },
            "description": "Client Error"
          }
        }
      },
      "delete": {
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "key",
            "in": "body",
            "required": true,
            "schema": {
              "type": "string",
              "minLength": 5,
              "maxLength": 50
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "type": "string"
            },
            "description": "Success"
          }
        }
      }
    }
  },
  "definitions": {},
  "parameters": {}
}
```

### DIFF
--- a/packages/adl/lib/openapi.ts
+++ b/packages/adl/lib/openapi.ts
@@ -631,9 +631,8 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
       case 'Boolean':
         return { type: 'boolean', enum: [adlType.value] };
       case 'Model':
-        // Is the type templated?
-        if (adlType.baseModels.length > 0 && !hasSchemaProperties(adlType.ownProperties)) {
-          // NOTE: Arbitrarily picking first type for now
+        // Is the type templated with only one type?
+        if (adlType.baseModels.length === 1 && !hasSchemaProperties(adlType.ownProperties)) {
           return mapADLTypeToOpenAPI(adlType.baseModels[0]);
         }
 
--- a/packages/adl/lib/openapi.ts
+++ b/packages/adl/lib/openapi.ts
@@ -2,7 +2,7 @@ import * as fs from 'fs';
 import * as path from 'path';
 import { Program } from '../compiler/program.js';
 import { ArrayType, InterfaceType, InterfaceTypeProperty, ModelType, ModelTypeProperty, Type, UnionType } from '../compiler/types.js';
-import { getDoc } from './decorators.js';
+import { getDoc, getFormat, getIntrinsicType, getMaxLength, getMinLength } from './decorators.js';
 import {
   basePathForResource,
   getHeaderFieldName,
@@ -73,7 +73,7 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
     // Write out the OpenAPI document to the output path
     fs.writeFileSync(
       path.resolve(options.outputFile),
-      JSON.stringify(root, null, 4));
+      JSON.stringify(root, null, 2));
   }
 
   function emitResource(resource: InterfaceType) {
@@ -231,7 +231,7 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
             }
             break;
           case 'content-type':
-            if (type.kind === "String" ){
+            if (type.kind === "String") {
               contentType = type.value;
             }
             break;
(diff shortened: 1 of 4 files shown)


packages/adl/lib/openapi.ts
```diff
@@ -603,6 +631,12 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
       case 'Boolean':
         return { type: 'boolean', enum: [adlType.value] };
       case 'Model':
+        // Is the type templated?
+        if (adlType.baseModels.length > 0 && !hasSchemaProperties(adlType.ownProperties)) {
+          // NOTE: Arbitrarily picking first type for now
```
daviwil, line 636: Do we want to have a rule about which type "wins" when a templated type has multiple base models?  Or just say that the rule is that this only works when there's a single type in the template list?

nguerrera: I may not understand all the ramifications, but seems like it would be safer to make this only work if there is just one base. I think more than one here would be something like `model Weird = string & int`. Am I following right? If so, that should probably be an error, but at least not treated as just string or just int.

nguerrera (approved):

packages/adl/lib/openapi.ts
bterlson: Better to be conservative for now. In general let's have a rule like "when in doubt, throw" 😁 It's good to avoid creating semantics people can take a dependency on that we might want to change later.

bterlson (approved):
Starting to feel useful, nice work!

packages/adl/lib/openapi.ts
daviwil: Sounds good! Widening the semantics later is definitely a safer approach :)

PR #245, opened 2021-01-15
Update grammar to reflect current state, use grammarkdown

### LINKED ISSUE #237: Grammar documentation
We had a language grammar defined for the prototype in an early proposal but neglected to keep it up to date. We should document the grammar as it is parsed now and in the future include grammar updates with any language proposals/changes.

We should use grammarkdown for this: https://www.npmjs.com/package/grammarkdown

nguerrera:
Fixes #237 by switching grammar to grammarkdown and updating it to reflect current state of the language.

First commit is separate, fixes up some remaining things that broke in VS Code after the code shuffle for debugging and building.

Grammarkdown is run during the build and will error out if the grammar is invalid in ways such as a reference to something that isn't defined. 

**EDIT**: The questions are now out of date with the evolution of the PR, which now follows the current RD parser implementation directly. Will follow up with separate issue(s) PRs on a couple of remaining design issues left over from discussing the questions below here and elsewhere. 

1. Triple-quoted string rules are tricky to specify and I just didn't get to it yet.

2. What is a MemberExpression intended for? Is this for future namespacing or for referencing members of model/interface properties (in decorator args maybe)? It's missing from the grammar now and I have just unqualified identifiers I am starting to wonder if we should consider having a distinct concept/naming of type nodes vs expression nodes so that we can allow for expressions that are not types in the future.~~~

3. Should we go ahead and prune out some redundant syntaxes that have fallen out of favor:
* C# attribute style `[ ... ]` for decorators?
* alias instead of model =?

4. I had trouble understanding why several things other than union and intersection are not primary expressions. I did not express the grammar matching that yet as I want to understand first. Is there a precedence issue that I am missing? TypeScript grammar I could find appears to have things like array type, tuple type, type reference as primary types, which is simpler to follow. I wrote the grammar that way for now until I understand this.


packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    `true`
+    `false`
+
+NumericLiteral :
+    DecimalLiteral
+    HexIntegerLiteral
+
+DecimalLiteral :
+    DecimalIntegerLiteral `.` DecimalDigits? ExponentPart?
+    `.` DecimalDigits ExponentPart?
```
bterlson, line 59: Is this critical? It adds some minor complexity to avoid having to start with `0.` which I think is good practice anyway...

bterlson: (Not PR feedback since you're just documenting, just something to think about simplifying).

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    one of `e` `E`
+
+SignedInteger :
+    DecimalDigits
+    `+` DecimalDigits
+    `-` DecimalDigits
+
+HexIntegerLiteral :
+    `0x` HexDigits
+    `0X` HexDigits
```
bterlson, line 85: Not needed.

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    HexDigit
+    HexDigits HexDigit
+
+HexDigit : 
+    one of `0` `1` `2` `3` `4` `5` `6` `7` `8` `9` `a` `b` `c` `d` `e` `f` `A` `B` `C` `D` `E` `F`
+
+StringLiteral :
+    `'` SingleStringCharacters? `'`
+    `"` DoubleStringCharacters? `"`
+    ``` BacktickStringCharacters? ```
```
bterlson, line 97: We should probably pick one of these at some point.

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+Punctuator : 
+    one of  `|` `:` `,` `;` `.` `<` `>` `(` `)` `{` `}` `[` `]` `@` `...`
+
+WhiteSpace :
+    <TAB>
+    <VT>
+    <FF>
+    <SP>
+    <NBSP>
+    <ZWNBSP>
```
bterlson, line 134: Interesting, is this both ZWSP and NBSP in one?

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    DecoratorList? `model` Identifier `{` ModelBody? `}`
+    DecoratorList? `model` Identifier `=` Expression `;`
+
+ModelBody :
+    ModelPropertyList `,`?
+    ModelPropertyList `;`?
+
+ModelPropertyList :
+    ModelProperty
+    ModelPropertyList `,` ModelPropertyList
```
bterlson, line 211:
```suggestion
    ModelPropertyList `,` ModelProperty
```

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    ModelPropertyList `;` ModelPropertyList
```
bterlson, line 212:
```suggestion
    ModelPropertyList `;` ModelProperty
```

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+    InterfacePropertyList `,` InterfaceProperty
+    InterfacePropertyList `;` InterfaceProperty
+
+InterfaceProperty :
+    DecoratorList? Identifier `(` ModelPropertyList? `)` `:` Expression
+
+Expression :
+    UnionOrIntersectionOrPrimaryExpression
+ 
+UnionOrIntersectionOrPrimaryExpression :
```
bterlson, line 240: I think the way expressions are done here won't scale if we add more expression types. Suggestion:

* Name the expression after the thing it will parse without recursing.
* The first RHS alternative is higher precedence productions

So our expression grammar might look like:

```
Expression : UnionExpression

UnionExpression :
    IntersectionExpression
    UnionExpression `|` IntersectionExpression

IntersectionExpression :
    PrimaryExpression
    IntersectionExpression `&` PrimaryExpression

... etc
```

I think this also makes the associativity of the operators more clear, personally.

bterlson (changes requested):
Nice work! Few suggestions.

packages/adl/language.grammar
nguerrera: I think so? This was copied from the example of JavaScript in grammarkdown README and matches ECMA-262. I took it wholesale because we also borrowed the identifier scanning from TypeScript so far. I must admit that I didn't pay careful attention to these codes, but this is https://www.fileformat.info/info/unicode/char/feff/index.htm

packages/adl/language.grammar
nguerrera: Fair point on scalability. The flip side is that it breaks my brain a little to have it possible for an expression without `|` in it match `UnionExpression`. I found it as I had it in the TypeScript spec doc at it was easier for me to grok this way.  This also touches on my question of whether the hierarchy in the parser is needed or if all the things other than Union and Intersection can be primary as in this PR....

What about something like this:

```
Expression : UnionExpressionOrHigher

UnionExpressionOrHigher :
    IntersectionExpressionOrHigher
    UnionExpression

UnionExpression:
    UnionExpressionOrHigher `|` IntersectionExpression

IntersectionExpressionOrHigher :
    PrimaryExpression
    IntersectionExpression

IntersectionExpression :
    IntersectionExpressionOrHigher `&` PrimaryExpression
```

packages/adl/language.grammar
```diff
@@ -0,0 +1,287 @@
...
+ExpressionList :
+    Expression
+    ExpressionList `,` Expression
+
+DecoratorList :
+    DecoratorList? Decorator
+
+Decorator :
+    `@` Identifier DecoratorArguments
+    `[` Identifier DecoratorArguments `]`
```
nguerrera, line 283: args are optional

packages/adl/language.grammar
nguerrera: You mean only allow lowercase x? I'm down but this is accurate to current scanner. I will do a separate PR to change that and update the grammar. I will batch it with some other cleanup I want to do after this exercise.

packages/adl/language.grammar
nguerrera: Agree. Do you have a preference?

packages/adl/language.grammar
nguerrera: Agreed on removing this. I'll incorporate the scanner + grammar change into a follow up.

packages/adl/language.grammar
bterlson: Yeah sorry, this was just another "I wish we were more conservative in the grammar" wishlist item 😁

packages/adl/language.grammar
bterlson: Double quote. I prefer single in my JS but double quotes are the standard IMO.

packages/adl/language.grammar
bterlson: Ahh, ZWNBSP/U+FEFF is the BOM and its use as an actual space is deprecated, should use ZWSP/U+200B instead. I think we can safely remove this. ADL source files MUST be UTF-8 😁

packages/adl/language.grammar
bterlson: I suppose I'm ok with this. I'm used to the ECMAScript grammar which uses the convention I suggested so I'm probably biased. It also matches more closely with the RD parser implementation.

packages/adl/language.grammar
nguerrera: Oh, yes, I see!    🤦‍♂️ 

I don't like so-called "UTF-8 with BOM" but a variety of code seems to create it underneath me. I'm fine with disallowing it but we should probably take care to have a good error and not just "unexpected character" or something because it can be really hard to spot. I'll remove it from the grammar. 

> ADL source files MUST be UTF-8

This makes me very happy. 😁

nguerrera, line 134:
```suggestion
```

bterlson:
Sorry I missed the questions before.

> What is a MemberExpression intended for? Is this for future namespacing or for referencing members of model/interface properties (in decorator args maybe)? It's missing from the grammar now and I have just unqualified identifiers I am starting to wonder if we should consider having a distinct concept/naming of type nodes vs expression nodes so that we can allow for expressions that are not types in the future.

Original prototypes by Garrett had member expressions in various scenarios, but we don't yet have need of them necessarily. You could imagine various use cases around referring to fields e.g. declaring a field as the type of another model's field, passing an individual field to a decorator, or when specifying a field to mutate in an as-yet uninvented feature for adding e.g. documentation to existing members.

> sould we go ahead and prune out some redundant syntaxes that have fallen out of favor

They won't be hard to add back in later, so sure let's do. If we want a uniform syntax for making an alias to a model or interface, we might want to bring alias back.

> I had trouble understanding why several things other than union and intersection are not primary expressions. I did not express the grammar matching that yet as I want to understand first. Is there a precedence issue that I am missing? TypeScript grammar I could find appears to have things like array type, tuple type, type reference as primary types, which is simpler to follow. I wrote the grammar that way for now until I understand this.

Which things in particular? With an example I can probably explain 😀

bterlson:
Ahh I see, TemplateExpression moved into PrimaryExpression. Are there others?

The thing that makes primary expressions special is that they don't have any worry about precedence because all its alternatives are equal/highest precedence.


This isn't the case for template application. Let's take the template application change. With this grammar, the something like `Foo.bar<T>` parses as `Foo.(bar<T>)` which is likely not what people would expect, instead expecting `(Foo.bar)<T>`.

Without member expressions or array expressions you wouldn't notice this, but the grammar still feels a little weird since identifier reference and template application are wildly different operations, so having two different syntax nodes for these seems convenient.

packages/adl/language.grammar
nguerrera: I'll do it the way you suggested, I think this is mostly just me not being accustomed to that convention. I'll get used to it. It's more concise and it scales easily. 😁

packages/adl/language.grammar
bterlson: Yeah you're right utf-8 doesn't say no BOM (UTF-8 with BOM is fine, even if it's not recommended and does nothing useful). I haven't experienced BOMs in my UTF-8 but if this is a thing then maybe we should keep it...

nguerrera: I think a lot of .NET code paths write out this way by default. 😪 It's (unfortunately) a thing. Putting it back.

VS Code:
image

nguerrera:
Doh, I missed https://github.com/Azure/adl/pull/245#issuecomment-764041053 until now. Sorry.

I moved template and array into primary (I remembered it being a bigger list, but it isn't). I've (temporarily) dropped member altogether as that was another question in my description. I don't know where you would use '.' today in practice. Template is also handled as optional template arguments on TypeReference.

I'm having trouble understanding where you use member in ADL today. One of the thing confusing me is that everything is called an "expression" but mostly all of our expressions so far are what I think of as types. Literals I can see as both. I think a member expression would be the first expression that is not a type if it is meant to refer to a type member. I think maybe decorators are the only place you could do that so far? I could also see it as referring to fully-qualified type, but we don't have namespaces yet.

I'm wondering if we should ultimately have Types and Expressions where every Type is an expression (but not vice versa)?

nguerrera:
I meant for Array to use PrimaryExpression. We're also missing `(` Expression `)` as a primary expression right now if you actually want `(a & b)[]`

bterlson:
>  I don't know where you would use '.' today in practice

If we adopt the namespaces rfc, you'll often be dotting things to get into namespaces. E.g.

```
namespace Foo {
  model Bar<T> { ... }
}

model Baz = Foo.Bar<T>;
```

>  I'm wondering if we should ultimately have Types and Expressions where every Type is an expression (but not vice versa)?

I don't follow this and the previous graf, can you say more/give an example?

nguerrera:
I thought since we have MemberExpression now but no namespaces yet that maybe Foo.bar would be to reference propery bar of model or interface Foo. In that case, I can't have a model property of type Foo.bar or an interface property with return type Foo.bar, but maybe I could use Foo.bar as a decorator arg? This would yield a case where we have a subset of expressions that can be types. Even if that is not the intent of member expression now, I could imagine in the far away future if we ever had some way to specify some glue implementation that we could have other such expressions (say addition) that are not types, so maybe it is worth calling these things types now rather than simply expressions?

nguerrera:
Chatted offline with Brian about this. Plan:

1. Sync grammar to current parser exactly in this PR.
2. File issue summarizing our discussion that ended in a couple of open design questions.

nguerrera:
@bterlson, I *think* this is now in sync with the RD parser and ready for another look. Haven't filed the follow-up issue yet.

bterlson (approved):
Nice work!

PR #246, opened 2021-01-20
Add initial ARM decorator library and types

daviwil:
This change contains an initial pass at implementing a library of ARM-focused decorators and types so that we can easily generate models and operations for RPaaS services.  I started from @markcowl's Confluent example files and implemented the `@ArmTrackedResource` decorator and its internal code expansion.

One thing I haven't implemented yet is the long-running operation decorators (`@slro`), we might need to discuss more about what that looks like in practice.

Check out the sample file: [confluent.adl](https://github.com/daviwil/adl/blob/arm-lib/packages/adl/samples/confluent/confluent.adl).  Expand the section below to see the OpenAPI output:

<details>
  <summary>OpenAPI output</summary>

```json
{
  "swagger": "2.0",
  "info": {
    "title": "(title)",
    "version": "0000-00-00"
  },
  "schemes": [
    "https"
  ],
  "paths": {
    "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Confluent/organizations/{name}": {
      "get": {
        "consumes": [],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "subscriptionId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resourceGroup",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/OrganizationModel"
            },
            "description": "Response envelope for ARM operations.",
            "headers": {
              "x-ms-client-request-id": {
                "type": "string"
              },
              "x-ms-correlation-request-id": {
                "type": "string"
              }
            }
          }
        }
      },
      "put": {
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "subscriptionId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resourceGroup",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resource",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/OrganizationModel"
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/OrganizationModel"
            },
            "description": "Response envelope for ARM operations.",
            "headers": {
              "x-ms-client-request-id": {
                "type": "string"
              },
              "x-ms-correlation-request-id": {
                "type": "string"
              }
            }
          }
        }
      },
      "patch": {
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "subscriptionId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resourceGroup",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resource",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/OrganizationModel"
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/OrganizationModel"
            },
            "description": "Response envelope for ARM operations.",
            "headers": {
              "x-ms-client-request-id": {
                "type": "string"
              },
              "x-ms-correlation-request-id": {
                "type": "string"
              }
            }
          }
        }
      },
      "delete": {
        "consumes": [],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "subscriptionId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resourceGroup",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "type": "object",
              "properties": {},
              "description": "Response envelope for ARM operations.",
              "x-adl-name": "{}"
            },
            "description": "Response envelope for ARM operations.",
            "headers": {
              "x-ms-client-request-id": {
                "type": "string"
              },
              "x-ms-correlation-request-id": {
                "type": "string"
              }
            }
          }
        }
      },
      "post": {
        "consumes": [],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "subscriptionId",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "resourceGroup",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/OrganizationKeys"
            },
            "description": "Response envelope for ARM operations.",
            "headers": {
              "x-ms-client-request-id": {
                "type": "string"
              },
              "x-ms-correlation-request-id": {
                "type": "string"
              }
            }
          }
        }
      }
    }
  },
  "definitions": {
    "OrganizationModel": {
      "type": "object",
      "properties": {
        "properties": {
          "$ref": "#/definitions/OrganizationProperties"
        }
      },
      "required": [
        "properties"
      ],
      "allOf": [
        {
          "$ref": "#/definitions/ArmTrackedResource"
        }
      ]
    },
    "OrganizationKeys": {
      "type": "object",
      "properties": {
        "primary": {
          "type": "string",
          "description": "The primary key."
        },
        "secondary": {
          "type": "string",
          "description": "The secondary Key."
        }
      },
      "required": [
        "primary",
        "secondary"
      ]
    },
    "OrganizationProperties": {
      "type": "object",
      "properties": {
        "createdTime": {
          "type": "string",
          "format": "date",
          "description": "UTC Time when Organization resource was created."
        },
        "organizationId": {
          "type": "string",
          "description": "Id of the Confluent organization."
        },
        "ssoUrl": {
          "type": "string",
          "description": "Single sign-on url for the Confluent organization."
        },
        "offerDetail": {
          "$ref": "#/definitions/OfferDetail"
        },
        "userDetail": {
          "$ref": "#/definitions/UserDetail"
        }
      },
      "description": "Details of the Confluent organization.",
      "required": [
        "createdTime",
        "organizationId",
        "ssoUrl",
        "offerDetail",
        "userDetail"
      ]
    },
    "ArmTrackedResource": {
      "type": "object",
      "properties": {
        "location": {
          "type": "string"
        },
        "tags": {
          "type": "object",
          "properties": {
            "k": {
              "type": "string"
            },
            "v": {
              "type": "string"
            }
          },
          "required": [
            "k",
            "v"
          ],
          "x-adl-name": "Map<string, string>"
        }
      },
      "description": "Common properties for all resources tracked by ARM.",
      "required": [
        "location"
      ],
      "allOf": [
        {
          "$ref": "#/definitions/ArmResource"
        }
      ]
    },
    "OfferDetail": {
      "type": "object",
      "properties": {
        "publisherId": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Id of the product publisher."
        },
        "id": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Id of the product offering."
        },
        "planId": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Id of the product offer plan."
        },
        "planName": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Name of the product offer plan."
        },
        "termUnit": {
          "type": "string",
          "maxLength": 25,
          "description": "Offer plan term unit."
        },
        "status": {
          "type": "string",
          "enum": [
            "Started",
            "PendingFulfillmentStart",
            "InProgress",
            "Subscribed",
            "Suspended",
            "Reinstated",
            "Succeeded",
            "Failed",
            "Unsubscribed",
            "Updating"
          ],
          "x-adl-name": "Started | PendingFulfillmentStart | InProgress | Subscribed | Suspended | Reinstated | Succeeded | Failed | Unsubscribed | Updating",
          "description": "SaaS offer status."
        }
      },
      "description": "Details of the product offering.",
      "required": [
        "publisherId",
        "id",
        "planId",
        "planName",
        "termUnit",
        "status"
      ]
    },
    "UserDetail": {
      "type": "object",
      "properties": {
        "firstName": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Subscriber first name."
        },
        "lastName": {
          "type": "string",
          "minLength": 5,
          "maxLength": 50,
          "description": "Subscriber last name."
        },
        "emailAddress": {
          "type": "string",
          "pattern": "\\w+@\\w+\\.\\w+",
          "description": "Subscriber email address."
        }
      },
      "description": "Details of the subscriber",
      "required": [
        "firstName",
        "lastName",
        "emailAddress"
      ]
    },
    "ArmResource": {
      "type": "object",
      "properties": {
        "id": {
          "type": "string"
        },
        "name": {
          "type": "string"
        },
        "type": {
          "type": "string"
        }
      },
      "description": "Common properties for all ARM resources.",
      "required": [
        "id",
        "name",
        "type"
      ]
    }
  },
  "parameters": {}
}
```

</details>

I'll merge the two `confluent.adl` example files after I get closer to having everything implemented.

**Questions**

- We can't have both a decorator and a model type called `ArmTrackedResource` so I chose to call the model `ArmTrackedResource` and the decorator `@TrackedResource`.  How do you feel about that?  Better suggestion?
- We can't have both an interface and a model named `Organization`.  Since the user typed in the name for `interface Organization {}`, I chose to generate the corresponding model named `OrganizationModel`.  Is there a better strategy here?  I'm considering adding a model-level decorator for renaming the type, what do you think?

### DIFF
--- a/packages/adl/lib/arm.ts
+++ b/packages/adl/lib/arm.ts
@@ -33,7 +33,7 @@ export function TrackedResource(
       // Create the resource model type and evaluate it
       const resourceModelNode =
         parseStatement<ModelStatementNode>(
-          `model ${target.name}Model { \
+          `model ${target.name}Resource{ \
             ... ArmTrackedResource; \
             properties: ${propertyType.name} \
           }`);
--- a/packages/adl/lib/arm.ts
+++ b/packages/adl/lib/arm.ts
@@ -1,0 +1,73 @@
+import { InterfaceStatementNode, ModelStatementNode, ModelType, Type, Statement } from "../compiler/types";
+import { Program } from "../compiler/program";
+import { parse } from "../compiler/parser.js";
+import { resource } from "./rest.js";
+
+function parseStatement<TNode extends Statement>(adlStatement: string): TNode {
+  // We're assuming that the parser will throw on bad input,
+  // so no error handling is added at this level
+  const script = parse(adlStatement);
+
+  if (script.statements.length === 0) {
+    throw new Error("No statements found in parsed input.");
+  }
+
+  return script.statements[0] as TNode;
+}
+
+// TODO: We can't name this ArmTrackedResource because that name
+//       is already taken.  Is this name sufficient?
+export function TrackedResource(
+  program: Program,
+  target: Type,
+  resourceRoot: string,
+  propertyType: Type) {
+
+  const checker = program.checker;
+  if (checker === undefined) {
+    throw new Error("Program does not have a checker assigned");
+  }
+
+  if (target.kind === "Interface") {
+    if (propertyType.kind === "Model") {
+      // Create the resource model type and evaluate it
+      const resourceModelNode =
+        parseStatement<ModelStatementNode>(
+          `model ${target.name}Model { \
+            ... ArmTrackedResource; \
+            properties: ${propertyType.name} \
+          }`);
+      const resourceModelType = checker.getTypeForNode(resourceModelNode) as ModelType;
+
+      // TODO: Add a standard API for this!
+      // Add the model type to the program's global scope
+      program.globalSymbols.set(resourceModelType.name, {
+        kind: 'type',
+        node: resourceModelNode
+      });
+
+      // Create a temporary interface type and parse it so
+      // that we can harvest its interface properties
+      const resourceInterfaceNode = parseStatement<InterfaceStatementNode>(
+        `interface Temp { \
+          @get get(@path subscriptionId: string, @path resourceGroup: string, @path name: string): ArmResponse<${resourceModelType.name}>; \
+          @put createOrUpdate(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelType.name}) : ArmResponse<${resourceModelType.name}>; \
+          @patch update(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelType.name}): ArmResponse<${resourceModelType.name}>; \
+          @_delete delete(@path subscriptionId: string, @path resourceGroup: string, @path name: string): ArmResponse; \
+        }`
+      );
+
+      // Add all of the properties from the parsed interface
+      for (const prop of resourceInterfaceNode.properties) {
+        target.properties.set(prop.id.sv, checker.checkInterfaceProperty(prop));
+      }
+
+      // Add the @resource decorator
+      resource(program, target, `/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/${resourceRoot}/{name}`);
+    } else {
+      throw new Error("TrackedResource property type must be a model");
+    }
+  } else {
+    throw new Error("TrackedResource decorator can only be applied to interfaces");
+  }
+}
--- a/packages/adl/lib/arm.ts
+++ b/packages/adl/lib/arm.ts
@@ -16,7 +16,8 @@ function parseStatement<TNode extends Statement>(adlStatement: string): TNode {
 }
 
 // TODO: We can't name this ArmTrackedResource because that name
-//       is already taken.  Is this name sufficient?
+//       is already taken.  Consider having decorators occupy a
+//       different namespace than types.
 export function TrackedResource(
   program: Program,
   target: Type,
@@ -31,25 +32,16 @@ export function TrackedResource(
   if (target.kind === "Interface") {
     if (propertyType.kind === "Model") {
       // Create the resource model type and evaluate it
-      const resourceModelNode =
-        parseStatement<ModelStatementNode>(
-          `model ${target.name}Resource = ArmTrackedResource<${propertyType.name}>;`);
-      const resourceModelType = checker.getTypeForNode(resourceModelNode) as ModelType;
-
-      // TODO: Add a standard API for this!
-      // Add the model type to the program's global scope
-      program.globalSymbols.set(resourceModelType.name, {
-        kind: 'type',
-        node: resourceModelNode
-      });
+      const resourceModelName = `${target.name}Resource`;
+      program.evalAdlScript(`model ${resourceModelName} = ArmTrackedResource<${propertyType.name}>;`);
 
       // Create a temporary interface type and parse it so
       // that we can harvest its interface properties
       const resourceInterfaceNode = parseStatement<InterfaceStatementNode>(
         `interface Temp { \
-          @get get(@path subscriptionId: string, @path resourceGroup: string, @path name: string): ArmResponse<${resourceModelType.name}>; \
-          @put createOrUpdate(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelType.name}) : ArmResponse<${resourceModelType.name}>; \
-          @patch update(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelType.name}): ArmResponse<${resourceModelType.name}>; \
+          @get get(@path subscriptionId: string, @path resourceGroup: string, @path name: string): ArmResponse<${resourceModelName}>; \
+          @put createOrUpdate(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelName}) : ArmResponse<${resourceModelName}>; \
+          @patch update(@path subscriptionId: string, @path resourceGroup: string, @path name: string, @body resource: ${resourceModelName}): ArmResponse<${resourceModelName}>; \
           @_delete delete(@path subscriptionId: string, @path resourceGroup: string, @path name: string): ArmResponse; \
         }`
       );
--- a/packages/adl/lib/arm.ts
+++ b/packages/adl/lib/arm.ts
@@ -33,10 +33,7 @@ export function TrackedResource(
       // Create the resource model type and evaluate it
       const resourceModelNode =
         parseStatement<ModelStatementNode>(
-          `model ${target.name}Resource{ \
-            ... ArmTrackedResource; \
-            properties: ${propertyType.name} \
-          }`);
+          `model ${target.name}Resource = ArmTrackedResource<${propertyType.name}>;`);
       const resourceModelType = checker.getTypeForNode(resourceModelNode) as ModelType;
 
       // TODO: Add a standard API for this!
--- a/packages/adl/samples/confluent/confluent.adl
+++ b/packages/adl/samples/confluent/confluent.adl
@@ -1,0 +1,90 @@
+@TrackedResource("Microsoft.Confluent/organizations", OrganizationProperties)
+interface Organization {
+  @post getKeys(@path subscriptionId: string, @path resourceGroup: string, @path name: string) : ArmResponse<OrganizationKeys>;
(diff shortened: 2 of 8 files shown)


packages/adl/lib/arm.ts
```diff
@@ -0,0 +1,73 @@
...
+        parseStatement<ModelStatementNode>(
+          `model ${target.name}Model { \
+            ... ArmTrackedResource; \
+            properties: ${propertyType.name} \
+          }`);
+      const resourceModelType = checker.getTypeForNode(resourceModelNode) as ModelType;
+
+      // TODO: Add a standard API for this!
+      // Add the model type to the program's global scope
+      program.globalSymbols.set(resourceModelType.name, {
```
daviwil, line 44: This is hacky, should we expose the program's `binder` and let that be the API for adding a global symbol?

packages/adl/samples/confluent/confluent.adl
```diff
@@ -0,0 +1,90 @@
+@TrackedResource("Microsoft.Confluent/organizations", OrganizationProperties)
+interface Organization {
+  @post getKeys(@path subscriptionId: string, @path resourceGroup: string, @path name: string) : ArmResponse<OrganizationKeys>;
+}
+
+@doc("Details of the Confluent organization.")
+model OrganizationProperties {
+  @doc("UTC Time when Organization resource was created.")
+  // @visibility('read')
```
daviwil, line 9: Visibility will be implemented soon

packages/adl/lib/arm.ts
bterlson: I wonder if we should have an `evalAdlScript` that creates bindings as normal?

bterlson: Also this could possibly be implemented more cleanly (from the ADL side) by having a library model template like `ResourceModel<T> { ... ArmTrackedResource; properties: T }`. Not the biggest of wins but it is probably a "best practice" (such that any exist at this point 😁) to have as many shapes in actual ADL as possible. If we had a handy function for instantiating a template that could be used instead of eval.

bterlson:
> We can't have both a decorator and a model type called ArmTrackedResource so I chose to call the model ArmTrackedResource and the decorator @TrackedResource. How do you feel about that? Better suggestion?

Seems fine to me... another thought is that decorators occupy a separate namespace, or looking at it another way, the `@` is part of the decorator's identifier (JS considered this at one point). I'm not sure that's needed though.


> We can't have both an interface and a model named Organization. Since the user typed in the name for interface Organization {}, I chose to generate the corresponding model named OrganizationModel. Is there a better strategy here? I'm considering adding a model-level decorator for renaming the type, what do you think?

I'm kind of sad about having a decorator to control generated names, I think a convention is probably better. Instead of OrganizationModel, maybe there is a better domain-specific name to pick, like OrganizationResource, or something?

packages/adl/lib/arm.ts
```diff
@@ -0,0 +1,73 @@
...
+
+// TODO: We can't name this ArmTrackedResource because that name
+//       is already taken.  Is this name sufficient?
+export function TrackedResource(
+  program: Program,
+  target: Type,
+  resourceRoot: string,
+  propertyType: Type) {
+
+  const checker = program.checker;
```
bterlson, line 26: I think it's impossible for a decorator to be invoked without the checker being defined fwiw, since the checker is what invokes the decorators.

packages/adl/lib/arm.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (checker === undefined) {
+    throw new Error("Program does not have a checker assigned");
+  }
+
+  if (target.kind === "Interface") {
```
bterlson, line 31: I think we should make use of type guards/asserts and have like `assertTypeIsInterface(target)`. Will be handy in many places, and gets narrowing without having to burn an indentation level 😁

bterlson (approved):
I'm getting excited.

packages/adl/lib/arm.ts
daviwil: Yep, sounds good to me!  I'll add some.

packages/adl/lib/arm.ts
daviwil: Good idea, I'll try that

packages/adl/lib/arm.ts
daviwil: > Seems fine to me... another thought is that decorators occupy a separate namespace, or looking at it another way, the `@` is part of the decorator's identifier (JS considered this at one point). I'm not sure that's needed though.
> 

I think it could totally make sense for decorators to have their own namespace since they are from a different "world" entirely.  Also very easy to identify which namespace we're looking in because of the `@`

> I'm kind of sad about having a decorator to control generated names, I think a convention is probably better. Instead of OrganizationModel, maybe there is a better domain-specific name to pick, like OrganizationResource, or something?

`OrganizationResource` sounds good to me!

nguerrera: Kind of aside, but I think it would be worth having a tracking issue to at some point down the maturity line draft a set of naming/design guidelines and then scrub through the early libs for compliance. I think it's really helpful when a language / ecosystem is born with some opinions on that.

packages/adl/lib/arm.ts
```diff
@@ -0,0 +1,73 @@
...
+      // TODO: Add a standard API for this!
+      // Add the model type to the program's global scope
+      program.globalSymbols.set(resourceModelType.name, {
+        kind: 'type',
+        node: resourceModelNode
+      });
+
+      // Create a temporary interface type and parse it so
+      // that we can harvest its interface properties
+      const resourceInterfaceNode = parseStatement<InterfaceStatementNode>(
```
nguerrera, line 51: Neat. This is making me want ADLisp even more 😄

nguerrera (approved):
Very cool. The ADL is looking much more readable than the resulting swagger already. Awesome to see this getting real.

daviwil:
@bterlson @nguerrera - What do you think of the latest commit, are the changes I made to the binder OK?

https://github.com/Azure/adl/pull/246/commits/6e7693598646a101d10e4709f003fdb4055733bd

I basically generalized things so that Program can handle binding at the time that source files (or script strings) get evaluated.

nguerrera:
Looks good to me.

bterlson:
I think it's fine, but a couple thoughts:

1. It may be nice to have loading and binding be separable phases (allows collecting all the source files in a program before the binding starts, which can be useful for throwing good error messages for example)
2. the public binder interface can know how to bind many things... another approach might be to add bindSourceFile that bindProgram calls, and move the logic for adding the global bindings into it.

But I think this is fine for now since it gets us wins 😁

daviwil:
Sounds good!  I'll be happy to do a clean up pass on all of this once this wave of demos is over.

PR #248, opened 2021-01-21
Add @secret and @visibility decorators, minor clean-up

daviwil:
This change adds the `@secret` and `@visibility` decorators to complete the set for the Confluent demo.  I'm using `@secret` here in the place of `SecureString` based on the discussion we had in Teams last week.  Also, `@visibility` information currently isn't getting used, I can try adding that before merging.

Also did some general cleanup in a couple places and rearranged the `confluent.adl` file to have a better flow for demos.

### DIFF
--- a/packages/adl/lib/decorators.ts
+++ b/packages/adl/lib/decorators.ts
@@ -83,7 +83,7 @@ export function minLength(program: Program, target: Type, minLength: number) {
       throw new Error("Cannot apply @minLength to a non-string type");
     }
   } else {
-    throw new Error("Cannot apply @format to anything that isn't a Model");
+    throw new Error("Cannot apply @minLength to anything that isn't a Model");
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/adl/lib/decorators.ts
```diff
@@ -104,10 +104,54 @@ export function maxLength(program: Program, target: Type, maxLength: number) {
...
+}
+
+export function isSecret(target: Type): boolean | undefined {
+  return secretTypes.get(target);
+}
+
+// -- @visibility decorator ---------------------
+
+export type VisibilityLevels = 'read' | 'write';
+const visibilityLevels = ['read', 'write'];
```
bterlson, line 139: Do we need to limit this? My thinking was literally any string is fine so you can just use whatever visibility flags you want.

nguerrera (approved):

packages/adl/lib/decorators.ts
nguerrera: Not at all blocking on this PR, but just a thought. This made me think of the extensible enum debates. Even if this is open-ended, if we had completion, it would be nice to hint at the common uses. More general question to capture somewhere: Should we have a way to supply typing for decorators so tooling could eventually help author them? I can file an issue.

bterlson: It's something we'll want eventually! In this case, though, I'm not sure how we can offer help really, assuming the idea of the visibility framework is "you get an arbitrary set of visibility flags". Presumably `@withVisibility` can offer completions based on the visibility flags actually present on the model its attached to, but unfortunately it won't help when you're typing the thing from scratch since you write the decorator before the model.

daviwil: I'll take out the value check for now, leaving it open for interpretation sounds good this early in the game!

daviwil:
Added AutoRest as a dependency of ADL so that the user doesn't have to know how to install it too.  Gonna go ahead and merge this now so that I can get another PR or two out from the changes I made yesterday, I can make any fixes you want on the AutoRest integration in the next ones!

PR #249, opened 2021-01-22
Add support for x-ms-pageable and arbitrary OpenAPI extensions

daviwil:
This change adds support for `x-ms-pageable` in the OpenAPI emitter via the `@pageable` decorator.  The idea is that the ADL author can mark an operation as being pageable and also identify the name of the response property containing the next link URL for paging purposes.  This is very specific to OpenAPI so we might need to consider coming up with a more general way to express this in the future.

I also added the `@extension()` decorator for attaching `x-*` extensions to elements of the Swagger document.  Currently I'm only applying them to model types, need to figure out the remaining places where they should go (I'll file an issue for that).  This decorator enabled me to add `"x-ms-azure-resource": true` to generated ARM resource model types.

Also including some cleanup I did to our example ADL files as a result of this, let me know if you want me to change any of that!

### DIFF
--- a/packages/adl/lib/arm.adl
+++ b/packages/adl/lib/arm.adl
@@ -19,7 +19,7 @@ model ArmTrackedResource<TProperties> {
 }
 
 @doc("Response envelope for ARM operations.")
-model ArmResponse<T> {
+model ArmResponseBase {
   @header statusCode: 200;
 
   @header("x-ms-client-request-id")
@@ -27,6 +27,17 @@ model ArmResponse<T> {
 
   @header("x-ms-correlation-request-id")
   correlationRequestId: string;
+}
 
+@doc("Typed response envelope for ARM operations.")
+model ArmResponse<T> {
+  ... ArmResponseBase;
   ... T;
+}
+
+@doc("Paged response")
+model Page<T> {
+  ... ArmResponseBase;
+  value: T[];
+  nextLink: string;
 }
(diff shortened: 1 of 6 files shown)


bterlson:
I think @johanste suggested we have a semantic list operation rather than explicit pageable, and I think I agree with that. Though this is fine for now, it's certainly easier.

I wonder if we should move pageable into adl core though? Many protocols have a notion of paging so maybe better to make it a general thing? That said, still need a way to provide nextlink which is a REST-specific (and maybe even azure-specific) thing.

nguerrera (approved):
Looks good!

daviwil:
We can definitely have a combo of a more generic `@list` and a `@nextLink` or `@pageable` that merely exists for giving the extra metadata for naming the `nextLink` property.  The OpenAPI emitter could default to `nextLink` when `@list` is encountered unless a `@nextLink` decorator overrides it.  What do you think @bterlson @johanste?

bterlson:
I think this aligns with what johan sketched here: https://github.com/Azure/adl/pull/223#discussion_r552878001. Except that I definitely prefer the idea of using a separate decorator to describe paging strategy (and maybe making it openapi-specific for now) rather than the parameter thing he put in the sketch.

daviwil:
Thanks, I'll make that tweak before merging!

johanste:
Sounds reasonable. `@list` is (the way I look at it) a protocol agnostic concept. And some lists may have (server driven) pagination. And we should have a "default projection" of this into Azure/REST. 

We should probably also include what client driven paging looks like.

daviwil:
@johanste Thanks for helping to keep us on track with having protocol agnostic concepts like this, I think it's critical to get it right early on!

PR #250, opened 2021-01-23
Prepare ADL for preview use, set up automatic npm publishing

### LINKED ISSUE #215: Set up npm publishing pipeline for ADL compiler so that it can be consumed by partners
We need to set up a publishing pipeline to deliver builds of the ADL compiler so that partners can try it out and provide feedback.

daviwil:
This change adds some minor extra polish so that we can automatically publish preview builds of the ADL compiler to npm as `@azure-tools/adl`.

We're going to try using the automatic release and version-bumping facilities [provided by Rush](https://rushjs.io/pages/maintainer/publishing/).  This means that for each PR we submit to this repo, we'll have to run `rush change` to generate a change entry file ([example here](https://github.com/daviwil/adl/blob/adl-packaging/common/changes/%40azure-tools/adl/adl-packaging_2021-01-23-00-23.json)) where we will describe what changed and what kind of version bump the changes require.  This will result in a few nice benefits:

- Automatic changelog generation
- Automatic version bumping for changed packages
- Automatic version bumping for packages that depend on changed packages, when necessary
- Only updated packages will be published to npm in CI runs

Once this PR gets merged, the automatic release process will begin!

Fixes #215.


packages/adl/package.json
```diff
@@ -1,13 +1,13 @@
...
+  "version": "0.1.300",
+  "description": "ADL Compiler Preview",
   "main": "dist/main.js",
   "bin": {
-    "adl": "dist/lib/cli.js"
+    "adl": "dist/compiler/cli.js"
   },
   "engines": {
-    "node": ">=10.12.0"
+    "node": ">=12.0.0"
```
nguerrera, line 10: I think elsewhere we said we could depend on node >= 14.

nguerrera (approved):

packages/adl/package.json
bterlson: I think in practice it does due to using esm support.

bterlson (approved):
Exciting! I will likely PR some tweaks to the readme later but this is great!

packages/adl/package.json
daviwil: I will update this to require 14, thanks for pointing it out!

PR #264, opened 2021-01-27
Commit auto-generated package and changelog changes

daviwil:
It turns out that the release automation that was used to publish `@azure-tools/adl@0.2.0` never actually pushed the `package.json` version bump and changelog changes back to `master` because we use a branch policy that prohibits it.  This means that we will now need to run `rush publish --apply` locally and then create a PR with the changes when we want to publish releases.  The good news is that the CI job on master can still attempt to publish any packages that need it!


nguerrera:
Should we consider adjusting branch policy?

nguerrera (approved):

daviwil:
> Should we consider adjusting branch policy?

It'd certainly make release automation easier, but I'm fine with following whatever the Azure SDK engineering guidelines are.

PR #265, opened 2021-01-27
Rework operator precedence

nguerrera:
NOTE: I'm not sure about this change, but I found it was easier to try it and put it up as a PR than to explain my intent with an issue description and no code. If we take this, there's room for further refactoring: keeping the checker working was particularly ~~hacky~~ opportunistic as I was mostly trying to see how the parser change would look.

As I was looking at proposing namespaces, I looked to consume MemberExpression and got confused in a way that I've been confused before, which is that parseMemberExpression() wouldn't necessarily parse a MemberExpression, it could parse something "higher". 

This then led me to seeing some issues like arrays of arrays not parsing and meaningless template applications like `7<9>` parsing but effectively crashing in the checker. It's debatable whether the latter needs to be a parse error or a semantic error, and even if it is a parse error, it is debatable if it needs to be a special rule or just excluded by grammar construction. But I found it not especially problematic to exclude it by grammar construction and that seems simplest to me. 

* Make it so that XxxExpression in grammar matches XxxExpression type in parser. Use
  XxxExpressionOrHigher in grammar to indicate XxxExpression or something with higher
  precedence. (I know we didn't agree the last time I raised this on the initial
  grammar PR, but I find this much easier to read and understand. You need less context
  when looking at an individual function or production IMHO.)

* Fix missing documentation of template parameters in grammar

* Fix arrays of arrays: `Foo[][]`

* Add support for parenthesized expressions to override precedence

* Combine member expressions and template applications. member expressions are now
  dotted identifier lists with optional trailing template arguments. (This is what another
  language might call a type reference expression, but we don't have a notion of "type"
  in our syntax as of yet. I think maybe we should, but I'm leaving that for a separate
  change and discussion. The renaming here would just obfuscate the separate issues
  I'm trying to address with this change.)

* Make array and member expressions primary expressions


packages/adl/compiler/parser.ts
```diff
@@ -277,7 +277,7 @@ export function parse(code: string) {
     }, pos);
 
     while (parseOptional(Kind.Bar)) {
-      const expr = parseIntersectionExpression();
+      const expr = parseIntersectionExpressionOrHigher();
```
bterlson, line 280: In an RD parser, I think "OrHigher" is always implied I believe (sometimes there won't be anything higher), but if others think this adds clarity I'm ok with it.

packages/adl/language.grammar
```diff
@@ -218,45 +225,57 @@ InterfaceProperty :
...
 
 TupleExpression :
     `[` ExpressionList `]`
 
+MemberExpression :
+    QualifiedIdentifier TemplateArguments?
+
+QualifiedIdentifier :
+    Identifier
+    QualifiedIdentifier `.` Identifier
```
bterlson, line 274: Depending on how we use `.` this could be expanded in the future (e.g. if it can be used to slice a property's type off an object you might want to do `Foo<bar>.prop`. Being conservative to start is fine though!

packages/adl/language.grammar
```diff
@@ -218,45 +225,57 @@ InterfaceProperty :
...
-    UnionExpression `|` IntersectionExpression
-
-IntersectionExpression :
-    ArrayExpression
-    IntersectionExpression `&` ArrayExpression
-
-ArrayExpression :
-    TemplateApplication
-    TemplateApplication `[` `]`
+    UnionExpressionOrHigher
```
bterlson, line 228: To me this change adds confusion, as now I need to trace through multiple productions to understand the associativity and precedence of an operator, and I don't really understand what the upside is. I've also not seen this convention used before so worry others would be confused too. But I'm not necessarily against this so if it helps, we can go with it.

packages/adl/compiler/parser.ts
nguerrera: Maybe because my limited parsing education didn't include languages with these sorts of potential ambiguities to tackle, it's very non-obvious to me. I can buy that it's just a part of the trade I haven't encountered yet, but I don't see the OrHigher taking away much, and it does make it clearer for those who wouldn't have learned and internalized this technique, like me. FWIW, I found XxxOrHigher in typescript's parser.ts.

bterlson: I can't say I understand TS's convention here either FWIW, as there are other cases that don't use a "OrHigher" suffix that recurse into other productions. Anyway, we can go with it!

packages/adl/language.grammar
nguerrera: The confusion that I had was that you have a thing called a UnionExpression in the compiler's syntax tree, which is not the same as a UnionExpression in the grammar. So for example, it seemed intuitive to me that when I wanted an a.b.c for using namespace  I call parseMemberExpression() to get it, but that function could give back something entirely different, e.g a string literal. The one place where we had parseMemberExpression was followed by a check: https://github.com/Azure/adl/pull/265/files#diff-1d84141f3f952bbb532f345c2ef53f9dc3b0d2055c8c30be3a9e8c2f16d20655L392

This breaks my brain. Call parseMemberExpression then check if we got a member expression.

packages/adl/compiler/parser.ts
```diff
@@ -419,42 +379,79 @@ export function parse(code: string) {
...
+        break;
+      case Kind.OpenParen:
+        expr = parseParenthesizedExpression();
+        break;
+      default:
+        throw error(`Unexpected token: ${Kind[token()]}`);
     }
 
-    throw error(`Unexpected token: ${Kind[token()]}`);
+    while (parseOptional(Kind.OpenBracket)) {
```
bterlson, line 438: I would factor this out into parseArrayExpression that starts with parsePrimaryExpression().

packages/adl/language.grammar
nguerrera: This bumps up against some other things I'm wondering about. Should there be one thing that represents all uses of dots or should a "member access" be a different thing than a "type reference" and maybe even use different syntax. I don't have an answer but it felt like being conservative now leaves us room to sort it out if/when we actually need to.

packages/adl/compiler/parser.ts
```diff
@@ -419,42 +379,79 @@ export function parse(code: string) {
     return args;
   }
 
-  function parseMemberExpression(): Types.Expression {
-    let base: Types.Expression = parsePrimaryExpression();
+  function parseMemberExpression(): Types.MemberExpressionNode {
```
bterlson, line 382: I could be wrong but doesn't this have the effect of parsing a lone identifier as a member expression (i.e. we don't use SyntaxKind.Identifier anymore)? If so, that seems undesirable as the two will have much different semantics (identifier does identifier resolution via scope walking, member expression does member lookup on its target)

nguerrera: It does. I guess I don't understand the problem yet. Naively, I'm seeing a dotted name as a generalized identifier. It qualifies an identifier to influence how it resolves. 

Using namespaces as we discussed having C# inspiration, I imagine this would compile:

```
namespace Foo {
    namespace Bar  {
        model Baz {}
     }
     model M = Bar.Baz;
}
```

So both the current context and the dot qualification are in play together on the same node.

I think the semantics would be different if the node represents a general `.` that can do more than qualify type names. Is there the same semantic problem without that?

I think if we are saying (for now maybe) that there is no other purpose for `.`, then "member expression" is no longer a good name here.

packages/adl/language.grammar
bterlson: Sorry, I'm referring to splitting productions like:

```
Expr :
    NextExpr
    Expr Op NextExpr
```

into

```
Expr :
    NextExpr
    OpExpr

OpExpr:
   Expr Op NextExpr
```

These are equivalent but the indirection adds a lot of complexity without giving much value from what I can tell?

bterlson: Regarding your point, it's often the case that it's easier and faster to parse a thing and check what you parsed and throw than to literally implement the grammar. And in fact, it is often easier to define the grammar like this too. ECMAScript uses "static semantics" to make specifying the grammar easier without having to flow through special contexts. For example, it's a syntax error to put a `yield` in an arrow function, but rather than create an entirely new syntax tree for this special case, the grammar parses it like a normal function body, and a static semantics says basically "if there's a yield expression in there anywhere, throw a syntax error". These two representations are equivalent so you can pick which to use.

In this case I'm in favor of the tightening for now so I think we're all good!

packages/adl/compiler/parser.ts
bterlson: or "parseArrayExpressionOrHigher" if you want to go with that convention.

packages/adl/language.grammar
nguerrera: I guess it's subjective, but I don't see a lot of complexity. For my brain, it's less complex this way.

* I find value in `Expr` in grammar mapping to `Expr` type in compiler's syntax tree rather than to `Expr | Next`.
* I also like that you are effectively documenting `&` being "higher" than `|` separately from them being things with an `|` and and `&` in the middle. This is the way these were taught to me in C, separately from where I was taught about RD parsers where the languages we used just didn't have these concerns. 

It's entirely possible that I'm just biased by my experience, and not necessarily in a long-term helpful way. 😅

packages/adl/compiler/parser.ts
bterlson: Regardless of what dotted identifiers mean, a lone identifier and a member expression are very different. You'll see this immediately when you fill in your checkMemberExpression function - the "if member.base !== undefined" case will be completely different from the else case. So they might as well be different nodes. What you have here works but it's kinda like if we had a single Expression node with an `op` field and optional RHS and LHS - represents the same thing, but in order to evaluate you have to poke at its fields to see what you've got rather than just looking at the node name.

packages/adl/language.grammar
bterlson: I think the shorter form corresponds to how the implementation works. If we have ExprOrHigher in the grammar, that corresponds to what parseExprOrHigher does in the compiler?

I think the concise form shows both associativity and precedence of the operator with fewer productions so I'd rather keep it that way unless you feel super strongly here!

nguerrera:
Closing this for now. There's a lot to change based on above and insightful offline conversation. I think it's going to be easier to evaluate a lot of it along with an actual implementation of accessing namespace members. Some of the smaller things could go in without it in minimal change form: arrays of arrays, parenthesized expressions. I'll queue that up if the namespace proposal feels like it might stall. Focusing on that for now.

nguerrera:
It actually ended up being easier to extract the minimal change to create a better baseline for the proposed grammar changes to add namespaces to follow. Putting up a PR now.

PR #267, opened 2021-01-28
Fix path resolution when loading decorator modules

daviwil:
This change fixes an issue encountered by @jamesc and @timotheeguerin where running ADL from a folder other than `packages/adl` will result in errors about a missing `lib` folder.  The issue is that we've been resolving decorator module paths relative to `process.cwd()` leading to ADL's standard library decorator files being sourced incorrectly.

The fix is to rework path resolution a bit so that we generate the absolute path of `lib` and `dist/lib` relative to `dist/compiler/program.js` wherever it's installed on the system so that our standard library files get loaded correctly.  I double checked that the `pet-store` decorator file still gets loaded correctly in this case.


packages/adl/compiler/program.ts
```diff
@@ -1,9 +1,12 @@
+import url from "url";
+import path from "path";
 import { readdir, readFile } from 'fs/promises';
 import { join } from 'path';
 import { createBinder, DecoratorSymbol, SymbolTable } from './binder.js';
 import { createChecker, MultiKeyMap } from './checker.js';
 import { CompilerOptions } from './options.js';
 import { parse } from './parser.js';
+import { resolvePath } from './util.js';
```
timotheeguerin, line 9: I see that;s how the other imports are too but what is the reason for having the `.js` extension, does feel weird when working with typescript

daviwil: Feels very weird!  We're using ECMAScript modules in this project which apparently require the extension

daviwil: My mistake, it looks like we're using commonjs modules.  Not sure where I came up with that...

bterlson: https://github.com/Azure/adl/blob/master/packages/adl/package.json#L61 we are a fancy module actually. And yeah, extension is required in that case, although I have forgotten why that was. TypeScript, at the time, didn't play super nicely with type: module projects.

timotheeguerin: Yeah seems like this is defined as es6 module but tsc renders it as a commonjs module. Might be good to stick to one or the other

timotheeguerin: nvm, tsconfig in adl/tsconfig override it back to ` "module": "esnext",`
https://github.com/Azure/adl/blob/452be501eb94806b75ea907dba34bc41bb185c05/packages/adl/tsconfig.json#L4

Issue #261, opened 2021-01-26
Convert "simple" autorest.testserver specs to ADL


daviwil:
Here's a set of easier testserver specs that we'll try to convert to ADL to see what gaps we find.  We aren't going to try to support every feature of Swagger 100% so we'll have to use judgement calls when deciding whether parts of these specs should be converted.

- [ ] body-string
- [ ] body-file
- [ ] body-byte
- [ ] body-integer
- [ ] body-number
- [ ] body-datetime
- [ ] body-time
- [ ] body-boolean

PR #268, opened 2021-01-28
Add body-boolean sample

timotheeguerin:
Adding body-boolean testserver sample (#261)

Thing that isn't possible right now is using `OkResponse<true>` or `OkResponse<false>` and `OkResponse<boolean | null>` can't be done without wrapping `boolean | null` into a model `NullableBoolean`


packages/adl/samples/testserver/body-boolean/body-boolean.adl
```diff
@@ -0,0 +1,36 @@
+
+@doc "Error"
+model Error {
+  code: int32;
+  message: string;
+}
+
+
+@doc "Success"
```
daviwil, line 9: I'll send a PR to add this to the ADL standard library, as soon as I do that this definition will give you an error saying that it's a duplicate.

packages/adl/samples/testserver/body-boolean/body-boolean.adl
```diff
@@ -0,0 +1,36 @@
...
+  @header statusCode: 200;
+  ... T;
+}
+
+
+@doc "AutoRest Bool Test Service"
+@resource "/bool"
+interface Bools {
+  @doc "Get true Boolean value"
+  @get getTrue(): true | Error;
```
daviwil, line 20: I think this should still be wrapped in `Ok<>` but we'll need to figure out a way to represent that without spreading.  @nguerrera @bterlson, any thoughts?

nguerrera: There’s a gist somewhere about returning primitives. @bterlson went through some options

nguerrera: I think one option was like a member of a model annotated as this is the body itself and the other members are for headers and decorators, don’t wrap this is an object. `@body` might have been the name. So `Ok<{ @body value: bool }>` or something like that?

timotheeguerin: Could maybe it be a new model `Response` which has a body parameter which is the generic value.
And then when parsing the return type of the operation do the case where it is either a raw model or of type Response

daviwil:
Can you also add this spec to the `regen-samples` script: https://github.com/Azure/adl/blob/master/packages/adl/scripts/regen-samples.js#L15

Once you do that, run `rushx regen-samples` and check in the output file that gets generated.

daviwil (approved):
Looks good!  I compared the generated file to the one in the testserver repo, all of the necessary bits are equivalent.

PR #270, opened 2021-01-28
Update rest and openapi libraries to support verb decorator subpaths

daviwil:
This change updates our `@verb` decorators for HTTP verbs to allow specifying subpaths for the `@resource` path of the parent interface.  Luckily the mechanism was already in place for this in the `openapi` library so I just needed to add the subpath metadata to the decorator functions and apply it.


nguerrera (approved):

daviwil:
Whoops, pushed hte latest comment to wrong PR!

PR #271, opened 2021-01-30
Adding in cognitive services management

markcowl:
Work in progress, shows an idealized modeling of the service, with notes about additions to ADL-RPaaS.

- Will add an alternative spec with ADL-RPaaS as currently defined.

### DIFF
--- a/packages/adl/samples/rpaas/cg.mgmt/README.md
+++ b/packages/adl/samples/rpaas/cg.mgmt/README.md
@@ -1,0 +1,48 @@
+## New ADL-RPaaS Elements
+- KeyCredential: Handles storage-like keys
+- @PrivatendpointProvider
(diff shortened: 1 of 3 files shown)


packages/adl/samples/rpaas/cg.mgmt/cognitive.vanilla.adl
```diff
@@ -0,0 +1,460 @@
...
+model ArmUpdateRequest<T>
+{
+    // note: placeholder, how do we access the value from an input type?
+    @header if-match: string = T.eTag;
+    ... ArmRequest<T>
+}
+
+model AccountResource: ArmTrackedResource<AccountProperties>;
+
+@ArmTrackedResource(AccountProperties, 'Microsoft.CognitiveServices/accounts')
```
daviwil, line 16: This is just `TrackedResource` in the ARM library right now, the name conflicts with the ArmTrackedResource type so I had to rename it.

PR #277, opened 2021-02-04
Add intrinsic null type, null unions, and emit x-nullable in OpenAPI

daviwil:
This change introduces an intrinisic `null` model type defined in `lib.adl` which is treated specially by the OpenAPI emitter.  I also expanded our union handling so that when a model type is encountered, we support unions with a single `null` type.  I don't believe we can deal with `oneOf` at all in Swagger, so I'm not allowing union types with arbitrary model types.

Note that I didn't add a `NullKeyword` or any type of parser-level support for the null type or value.  I did try to add this but found that the parser and checker don't have a way to deal with a literal type with the same symbol as its type (`null` and `null`).  I found that the current approach was more straightforward to get `null` support off the ground but I can go add more language-level support if you think it's necessary!

### DIFF
--- a/packages/adl/lib/openapi.ts
+++ b/packages/adl/lib/openapi.ts
@@ -483,7 +483,13 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
 
   function getSchemaForUnion(union: UnionType) {
     let type: string;
-    const kind = union.options[0].kind;
+    const nonNullOptions = union.options.filter(t => !isNullType(t));
+    const nullable = union.options.length != nonNullOptions.length;
+    if (nonNullOptions.length === 0) {
+      throw new Error("Cannot have a union containing only null types.");
+    }
+
+    const kind = nonNullOptions[0].kind;
     switch (kind) {
       case 'String':
         type = 'string';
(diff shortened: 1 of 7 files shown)


daviwil:
Here's an example Swagger output for `samples/nullable/nullable.adl`:

```
{
  "swagger": "2.0",
  "info": {
    "title": "(title)",
    "version": "0000-00-00"
  },
  "schemes": [
    "https"
  ],
  "paths": {
    "test/get": {
      "get": {
        "operationId": "NullableMethods_get",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "name": "someParam",
            "in": "query",
            "required": true,
            "type": "string",
            "x-nullable": true
          },
          {
            "name": "modelOrNull",
            "in": "body",
            "required": true,
            "schema": {
              "type": "object",
              "properties": {
                "num": {
                  "type": "integer",
                  "format": "int32"
                }
              },
              "required": [
                "num"
              ],
              "x-nullable": true,
              "x-adl-name": "AnotherModel | null"
            }
          }
        ],
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/HasNullables"
            },
            "description": ""
          }
        }
      }
    }
  },
  "definitions": {
    "HasNullables": {
      "type": "object",
      "properties": {
        "str": {
          "type": "string"
        },
        "strOrNull": {
          "type": "string",
          "x-nullable": true,
          "x-adl-name": "string | null"
        },
        "modelOrNull": {
          "type": "object",
          "properties": {
            "num": {
              "type": "integer",
              "format": "int32"
            }
          },
          "required": [
            "num"
          ],
          "x-nullable": true,
          "x-adl-name": "AnotherModel | null"
        },
        "literalsOrNull": {
          "type": "string",
          "enum": [
            "one",
            "two"
          ],
          "x-nullable": true,
          "x-adl-name": "one | two | null"
        }
      },
      "required": [
        "str",
        "strOrNull",
        "modelOrNull",
        "literalsOrNull"
      ]
    }
  },
  "parameters": {}
}
```

packages/adl/lib/openapi.ts
```diff
@@ -494,19 +494,52 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
...
+        break;
       default:
         throw invalidUnionForOpenAPIV2();
     }
 
     const values = [];
+    if (type === 'model') {
+      // Model unions can only ever be a model type with 'null'
+      if (union.options.length > 2) {
+        throw invalidUnionForOpenAPIV2();
```
nguerrera, line 508: Nit: The error message here is no longer fully accurate. 

Maybe we should have two different errors or allude to `| null` as another valid possibility in the single error?

nguerrera (approved):

packages/adl/lib/openapi.ts
daviwil: I'll add another message!

packages/adl/lib/openapi.ts
```diff
@@ -494,19 +494,52 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
...
+      }
+
+      const otherType = union.options[1];
```
nguerrera, line 511: do we allow `null | Model` order? Should we?

nguerrera: Thinking about it some more, I think we should allow it.  I would ever use that style, but it's good to keep `A | B === B  | A` as an invariant. I guess we should also allow `Model | null | null` too along the same theoretical lines. Not at all urgent though.

daviwil: I'll see if I can add it really quickly

daviwil: Done, code's a bit simpler now too: https://github.com/Azure/adl/pull/277/commits/f43df17f65cd147b3af65e47ca179d5cca09621e

timotheeguerin (approved):

daviwil:
Going to merge this now, if anyone has any further comments I'll be happy to address them in a subsequent PR!

bterlson:
My only comment is this looks more reasonable than I expected it to.

PR #266, opened 2021-01-28
Add converted body-complex.adl from testserver specs, other improvements

daviwil:
[Creating this as a draft PR first, will expand description when ready for review]

PR #278, opened 2021-02-04
Add script and PR step to check sample output for regressions

daviwil:
I harvested this script and CI check frm #266 since we need it sooner rather than later.  Basically, it checks the Swagger emitter output for a specific list of sample folders to see if there are any changes in output compared to a checked-in set of output files.

When we make any changes that affect emitter output, we'll need to run `npm regen-samples` and check in the output files if we're satisfied with the changes that we see.

### DIFF
--- a/packages/adl/package.json
+++ b/packages/adl/package.json
@@ -16,7 +16,8 @@
     "watch": "tsc -p . --watch",
     "prepare": "npm run build",
     "test": "mocha 'dist/test/**/*.js'",
-    "grammargen": "grammarkdown --newLine LF language.grammar"
+    "grammargen": "grammarkdown --newLine LF language.grammar",
+    "regen-samples": "node scripts/regen-samples.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


packages/adl/package.json
```diff
@@ -16,7 +16,8 @@
     "watch": "tsc -p . --watch",
     "prepare": "npm run build",
     "test": "mocha 'dist/test/**/*.js'",
-    "grammargen": "grammarkdown --newLine LF language.grammar"
+    "grammargen": "grammarkdown language.grammar",
```
nguerrera, line 19: (Still miffed that --newLine didn't actually solve the problem. I have a note to tweak .gitattributes instead, so removing it here is fine, it doesn't seem to do anything.)

daviwil: Oh, I think I may have merged this wrong, I'll add that back.

nguerrera (approved):
Nice, this will be a lot easier than doing one-off manual checks on changes in the output and make it much easier to review!

packages/adl/package.json
nguerrera: OK, I added it to fix something and it didn't achieve anything and I still have the problem. But probably better not to pollute the history. I'll remove it with a different fix in another PR. Thanks.

PR #279, opened 2021-02-05
Add standard HTTP response types to rest.adl

daviwil:
Moving the `Ok<T>` type to be a standard definition in `rest.adl`.  Anything else I should add?

### DIFF
--- a/packages/adl/lib/rest.adl
+++ b/packages/adl/lib/rest.adl
@@ -2,3 +2,32 @@ model Ok<T> {
   @header statusCode: 200;
   ... T;
 }
+
+model Created {
+  @header statusCode: 201;
+}
+
+model Accepted {
+  @header statusCode: 202;
+}
+
+model Moved {
+  @header statusCode: 301;
+  @header location: string;
+}
+
+model NotModified {
+  @header statusCode: 304;
+}
+
+model Unauthorized {
+  @header statusCode: 401;
+}
+
+model NotFound {
+  @header statusCode: 404;
+}
+
+model Conflict {
+  @header statusCode: 409;
+}
--- a/packages/adl/lib/rest.adl
+++ b/packages/adl/lib/rest.adl
@@ -3,8 +3,9 @@ model Ok<T> {
   ... T;
 }
 
-model Created {
+model Created<T> {
   @header statusCode: 201;
+  ... T;
 }
 
 model Accepted {
--- a/packages/adl/lib/rest.adl
+++ b/packages/adl/lib/rest.adl
@@ -1,34 +1,37 @@
-model Ok<T> {
+model OkResponse<T> {
   @header statusCode: 200;
   ... T;
 }
 
-model Created<T> {
-  @header statusCode: 201;
-  ... T;
-}
-
-model Accepted {
-  @header statusCode: 202;
-}
-
-model Moved {
-  @header statusCode: 301;
+model LocationHeader {
   @header location: string;
 }
 
-model NotModified {
+model CreatedResponse {
+  @header statusCode: 201;
+}
+
+model AcceptedResponse {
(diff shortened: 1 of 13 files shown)


packages/adl/lib/rest.adl
```diff
@@ -0,0 +1,5 @@
+@doc "Success"
```
bterlson, line 1: Probably don't need to add docs for folks.

daviwil: Sounds good!

bterlson:
I would probably add at least some common response statuses, maybe:

* Created (201)
* Accepted (202)
* Moved (301)
* NotModified (304)
* Unauthorized (401)
* NotFound (404)
* Conflict (409)

Could even argue adding all of them but, maybe not for now...

daviwil:
Certainly happy to add all of those.  Any in particular that would need to be templated?  Maybe just `Created<T>`?

bterlson:
Yeah I think just created. I think we'll want default template params eventually because even `Ok` might not want to be templated if you just want to spread it in to some model type somewhere.

nguerrera (approved):

daviwil:
Should I add an optional `location` header property to `Created`?  It might be populated in some cases instead of returning a body.  Perhaps that should be a separate type?

bterlson:
It feels separate to me, personally. I thought looking at [how I used it before](https://gist.github.com/bterlson/1f1a569e4a6d7a8e93b2434bbc123238) might help but it looks like I didn't even use a separate `Created` abstraction.

daviwil:
Your file does have a `CreatedResponseWithLocation` but it still keeps the `<T>` which seems wrong unless this is intended to encapsulate the entire polling operation until the final resource is returned.

daviwil:
Also, should we have the `Response` suffix?  Definitely makes things a little clearer despite not being as terse.

bterlson:
I seem to recall you could get some response body along with a location, but I don't recall details. I agree it seems weird.

I wouldn't object to Response suffixes if you prefer them!

daviwil:
Hmmm, yeah, I suppose a `CreatedResponseWithLocation` could return a different type of body with more information.  OK, I'll add a separate one for that!

daviwil:
@nguerrera What do you think about the response models having `-Response` suffixes?  I like how `Ok<T>` looks but `OkResponse<T>` is less ambiguous.

nguerrera:
I like the Response suffix. Over time it will be goodness to build up some conventions like that. I’m generally not opposed to a bit of verbosity to make things clearer.

bterlson:
For the record I don't think CreatedResponseWithLocation need be a built-in? There seems like too many permutations of headers you might want to make built-ins for all of them.

bterlson:
Ok just thinking through how these types would get used, I get the feeling that the built-in types would be most useful as untemplated models (or models with default, empty template parameters). Users would use templates to conveniently fill in their model types along with their envelope type. So it might be like:

```
model BrianServiceOkResponse<T> {
  ... OkResponse;
  ... ETagHeader;
  ... T;
}

model BrianServiceCreatedResponse {
  ... CreatedResponse;
  ... ETagHeader;
  ... CreatedHeader;
}
```

nguerrera:
Default T={} seems interesting as you could do above or use OkResponse<T> directly as now if you have nothing else to add. I find that nice for little samples at least. Of course we’d have to add the feature.

daviwil:
@bterlson I like that approach!  That seems more idiomatic to ADL.  I'll make changes for that tomorrow.

daviwil:
OK, I've updated the model types to be a litlte more composable, but I left `OkResponse<T>` as templated since it will most commonly be used that way.  What do you think?

bterlson (approved):

Issue #234, opened 2021-01-12
Namespace design


nguerrera:
Today we have interfaces which are bags of operations, but there is no hierarchy to them. 

This could be augmented or replaced with namespaces which could be hierarchical as they are in other languages.

The hierarchy of namespaces could align with the path hierarchy for REST API definition.

For example, GET /foo/bar/baz could something like bet

```
namespace foo {  
  namespace bar {
     @get baz():string
  }
}
```

### Some design points to consider:

* Namespaces should probably be "partial" like in other languages, meaning you can spread its member namespaces, operations across more than one source file.

* Should models go inside namespaces too?

* Do we need to retain interfaces at all or is it just that namespaces can have functions?

* How do namespaces interact with import/export/package management (TODO: issue link)

nguerrera:
Notes from chat with @bterlson:
* implicit foo/bar/baz is "cute", go with @uri (name maybe TBD) + path on each namespace

daviwil:
So far I like this a lot!  Here are some general thoughts:

- How will namespaces get "sourced" from extension packages?  If we're using a plain `using Some.Namespace`, do we assume that all of the extension packages are pulled in and added to a "load path" of sorts?

- Can we be more explicit about pulling in JS files to import decorator code into a specific namespace?  Right now we're just pulling in whatever .js files are in the folder path, but maybe it's better to be more explicit?  `import` could possibly be used for this:  `import './decorators.js'`.  If you use this statement inside of a sub-namespace, the imported decorators get added there.

One interesting aspect of this (which I've seen discussed in other issues here) is that we can use the root namespace to determine the service title and metadata when emitting Swagger:

```
@title("My Service Client")
namespace MyServiceClient {
}
```

Now the ADL user can refer to a specific namespace when compiling the output:

```
adl compile --namespace MyServiceClient samples/myserviceclient
```

nguerrera:
> How will namespaces get "sourced" from extension packages? If we're using a plain using Some.Namespace, do we assume that all of the extension packages are pulled in and added to a "load path" of sorts?

I don't know, except that at a high level somewhere I think you specify which packages (and versions!) you are referencing and somehow the compiler gets enough info to get to their source make their definitions available as described here. I think ironing out the mechanics is up to #236 to define, but not saying package names in the source is something that @bterslon and I discussed and we think it will be more productive. C# also works this way. That said, I suspect we will see a strong correspondence between packages and namespaces as you see in .NET.

[...]

PR #280, opened 2021-02-08
Rename 'interface' concept to 'namespace'

daviwil:
This is the first step of implementing @nguerrera's namepsace proposal (#234), basically just renaming the `interface` concept to `namespace`.  I haven't changed any of the mechanics for how this works, I think we could try that in a separate PR if you think it's safer to do it that way.


daviwil:
@nguerrera If you'd like, I can exclude the grammar file changes from this PR so that it doesn't conflict with #274!

nguerrera (approved):

daviwil:
Thanks!  I'll go ahead and merge this.

PR #282, opened 2021-02-08
Add back line and column tracking

nguerrera:


packages/adl/compiler/scanner.ts
```diff
@@ -102,7 +103,12 @@ const enum TokenFlags {
   TripleQuoted = 1 << 2,
 }
 
-export function createScanner(input: string, onError = throwOnError) {
+export function createScanner(source: string | SourceFile, onError = throwOnError): Scanner {
+  if (typeof source === 'string') {
+    source = createSourceFile(source, '<anonymous file>');
+  }
```
nguerrera, line 109: This is temporary to get back to parity before my refactor with the smallest change as quickly as possible. Next step is to actually have the compiler create the source files with their paths and pass them in.

packages/adl/compiler/scanner.ts
```diff
@@ -148,8 +155,7 @@ export function createScanner(input: string, onError = throwOnError) {
           if (lookAhead(1) === CharacterCodes.lineFeed) {
             position++;
           }
-          return next(Token.NewLine);
-
+          // fallthrough
```
nguerrera, line 158: Felt better in new code to use a fallthrough so went back and did same elsewhere.

packages/adl/compiler/parser.ts
```diff
@@ -587,7 +587,8 @@ export function parse(code: string) {
   }
 
   function error(msg: string) {
-    throw new Error(msg);
+    const pos = scanner.source.getLineAndCharacterOfPosition(scanner.tokenPosition);
+    throw new Error(`[${pos.line + 1}, ${pos.character + 1}] ${msg}`);
```
nguerrera, line 591: This restores the error to before my refactoring. `scanner.source.path` should be used here, but it's "<anonymous file>" right now. Working on plumbing that through next. I wanted to get back to parity with pre-refactoring ASAP.

bterlson (approved):

daviwil (approved):

PR #283, opened 2021-02-09
Remove some casts

nguerrera:
It occurred to me while reading the code that we could avoid a bunch of casts if `Node` followed the same pattern as `Type` where it is a union and the common properties go to `BaseNode` interface. Am I missing a reason why we shouldn't do this?


common/changes/@azure-tools/adl/remove-some-casts_2021-02-09-16-52.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@azure-tools/adl",
+      "comment": "Remove some casts from source code",
```
nguerrera, line 5: Is there a way to say: 'This change is a no-op for customers'. I wouldn't put cleanup like this in release notes at all.

daviwil: When you use `rush change`, just press Enter to say "no changes" and it will create an empty change file which doesn't cause a version bump.

packages/adl/compiler/types.ts
```diff
@@ -126,14 +126,26 @@ export enum SyntaxKind {
...
-export interface Node {
+export interface BaseNode {
   kind: SyntaxKind;
   pos: number;
   end: number;
   parent?: Node;
 }
 
-export interface ADLScriptNode extends Node {
+export type Node =
```
daviwil, line 136: Alternatively this could be `NodeType` and the base class could stay `Node`, though I suppose that might result in even more code changes updating the signatures.  Not a big deal either way!

daviwil (approved):

common/changes/@azure-tools/adl/remove-some-casts_2021-02-09-16-52.json
nguerrera: Ah! thanks!

nguerrera: I see now it tells me this, and I'm rather embarrassed that I asked. 😊

packages/adl/compiler/types.ts
nguerrera: Yes, it's not a big deal either way. I think it works better to use the good name for the union type as that is the type you want to reference everywhere but `extends` as I have it and also as `Type`/`BaseType` is already.

nguerrera:
@bterlson, I'll wait for your review on this one.

bterlson (approved):
Seems good to me. I suspect the reason the TS compiler doesn't use this pattern is compilation time, but since we're a few orders of magnitude smaller that is simply not a worry. I like this.

PR #290, opened 2021-02-10
Fix resolution of declared parameters in resource sub-paths

daviwil:
This change fixes an issue that @jamesc found, demonstrated by this ADL snippet:

```
@resource("/v1")
namespace LibraryService {
  // TODO: This path should be "/v1/shelves/{name}:merge" but doesn't work in ADL right now
  @post("shelves/{name}:merge")
  op MergeShelves(... MergeShelvesRequest): Shelf;
}
```

The resulting Swagger has an additional `{name}` appended in the operation path because the `@post` subpath's `{name}` parameter isn't detected:

```
  "paths": {
    "/v1/shelves/{name}:merge/{name}": {
      "post": {
```

The fix is to build the full route path before attempting to detect path parameters.  The resulting Swagger looks like this:

```
  "paths": {
    "/v1/shelves/{name}:merge": {
      "post": {
```

### DIFF
--- a/packages/adl/lib/openapi.ts
+++ b/packages/adl/lib/openapi.ts
@@ -114,11 +114,24 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
   /**
    * Translates endpoint names like `read` to REST verbs like `get`.
    */
-  function pathForEndpoint(prop: NamespaceProperty, pathParams: ModelTypeProperty[], declaredPathParamNames: string[]): [string, string[], string?] {
+  function pathForEndpoint(prop: NamespaceProperty, pathParams: ModelTypeProperty[]): [string, string[], string] {
     const paramByName = new Map(pathParams.map((p) => [p.name, p]));
-    const pathSegments = [];
+    const route = getOperationRoute(prop);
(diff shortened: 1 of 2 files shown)


packages/adl/lib/openapi.ts
```diff
@@ -114,11 +114,24 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
...
+    const route = getOperationRoute(prop);
+    const inferredVerb = verbForEndpoint(prop.name);
+    const verb = route?.verb || inferredVerb || 'get';
+
+    // Build the full route path including any sub-path
+    const routePath =
+      (currentBasePath || "") +
+      (route?.subPath
+       ? `/${route?.subPath?.replace(/^\//g, '')}`
+       : (!inferredVerb && !route ? "/get" : ""));
```
daviwil, line 128: I admit, this is a little gnarly, so I can break it up a bit with some intermediate assignments if you like.

bterlson: could probably be abstracted into a utility function too

bterlson (approved):

nguerrera (approved):

PR #291, opened 2021-02-11
Implement #276 composition reform

### LINKED ISSUE #233: Add visibility and withVisibility decorators to ADL standard library
- [x] `@visibility`: Defines the visibility level of a given model property
- [ ] `@withVisibility`: Filters the properties of a given type based on a particular visibility level

### LINKED ISSUE #276: Composition reform
Implementing the visibility framework proposed in RPaaS requires some careful thought. Today ADL uses `...` to mean, effectively, `extends` - properties from the parent models copied into the child model but preserve their original identity. If we combine this semantics with decorators, we observe the following:

```
model A { x: string }

@makePropertiesOptional
model B { ... A }
```

The only way to make sense of this is that the `makePropertiesReadOnly` decorator can't do anything. It doesn't see any own properties on B, and it can't walk up to A to change those properties because that would be like changing A itself. Probably not what you want.

This has two problems, as I see it: it would be useful to compose objects from other objects and apply decorators to the result - this gives "views". Additionally, the `spread` syntax is really creating a subtype but that is sort of non-obvious from the syntax.

A way to fix this is the following:

## Spread creates new properties

The spread syntax creates new properties in the parent by re-evaluating the properties on the target in the context of the parent. So, the following two declarations are roughly identical:

```
@md model A { @pd x: string }
@md model B { ... A }
```

Note how model B needs the `@md` decorator - `...` peels the properties out of its target, but does nothing with its container. With this we can implement "views" and the visibility framework:

```js
const visibility = new Map();
export function getVisibility(entity) {
  return visibility.get(entity);
}
export function v(c, t, v) {
  visibility.set(t, v);
}

export function wv(c, t, v) {
  const filter = (_, prop) => {
    const vis = getVisibility(prop);
    return vis && vis !== v;
  }

  mapFilterOut(t.properties, filter);
}

function mapFilterOut(map, pred) {
  for (const [key, prop] of map) {
    if (pred(key, prop)) {
      map.delete(key);
    }
  }
}
```

```
model Person {
  @v("read") id: string;
  @v("write") secret: string;
  name: string;
}

@wv("read")
model Readable<T> {
  ... T;
}

@wv("write")
model Writable<T> {
  ... T;
}

@resource("/hello")
interface Hello {
  @get read(@path id: string): Readable<Person>;
  @post create(person: Writable<Person>): Readable<Person>;
}
```

Here we use model templates and decorators together to give us nice Readable and Writable templates. We don't even need to declare the usages which is super handy.

## Introduce `model extends`

Unfortunately, approving the above makes it very difficult to recover the "inheritance hierarchy" when e.g. emitting to OpenAPI with it's `allOf` semantics. So I propose we make extensions explicit using `model extends` syntax:

```
model Person {
  name: string;
}

@childDecorator
model Child extends Person {
  toys: string[];
}
```

In this model, the child decorator sees only the `toys` property. It inherits the `name` property from Person, and cannot modify it or add metadata too it. This means something like this is nonsense:

```
model Person {
  @v('read') id: string;
  name: string;
}

@wv('read')
model Child extends Person {
  toys: string[];
}
```

The `wv` decorator will only see the toys property. The child model will still inherit the full Person model. This could be worked around with a small change to explicitly inherit from the readable version of Person:

```
@wv('read')
model Child extends Readable<Person> {
  toys: string[];
}
```

bterlson:
This PR implements the composition reform discussed in #276. An example leveraging these changes has been added to samples/visibility, please take a look!

The general idea in #276 stands, with the exception that properties created by creating a new model and copying properties (i.e. intersection and spread operations) keep track of the property they copied from. Emitters that care where a property came from can walk this linked list of parents. This is used in the openapi emitter to find a canonical definition for a property for ref deduping.

(This could use some thought, however - it is conceivable that a spread target model has decorators on it not present in the spread source that alter the semantics of the property, and so it would not be correct to dedupe it against a source property definition. But it's useful so I've left it in for now.)

I have yet to update the existing samples to reflect these changes, but I put the diffs in so you can see the effect of this change on existing ADL.

Closes #276. Closes #233.


packages/adl/compiler/checker.ts
```diff
@@ -1,3 +1,4 @@
+import { options } from 'yargs';
```
bterlson, line 1: ugh vscode auto-import hosed me good.

packages/adl/compiler/checker.ts
```diff
@@ -147,10 +149,6 @@ export function createChecker(program: Program) {
       // template instantiation
       const args = model.templateArguments.map(getTypeName);
       return `${model.name}<${args.join(', ')}>`;
-    } else if (model.ownProperties.size == 0 && model.baseModels.length > 0) {
```
bterlson, line 150: I think we lose good names for intersections. I'm ok with this.

packages/adl/compiler/checker.ts
```diff
@@ -394,6 +416,55 @@ export function createChecker(program: Program) {
...
+          sourceProperty: prop
+        }, true)
+        props.push(newProp);
+      }
+    }
+
+    return props;
+  }
+
+  function* walkPropertiesInherited(model: ModelType) {
```
bterlson, line 454: This could be exported, it's pretty handy for decorators I'd guess.

packages/adl/compiler/checker.ts
```diff
@@ -414,25 +485,11 @@ export function createChecker(program: Program) {
       });
     }
   }
-
-  function createModelType(typeDef: Omit<ModelType, 'properties'>): ModelType {
```
bterlson, line 418: Don't need this special method anymore since we don't do any special own/derived property book keeping for models.

packages/adl/compiler/checker.ts
```diff
@@ -394,6 +416,55 @@ export function createChecker(program: Program) {
...
+      if (targetType.kind !== 'Model') {
+        throw new Error('Cannot spread properties of non-model type.');
+      }
+
+      // copy each property
+      for (const prop of walkPropertiesInherited(targetType)) {
+        const newProp = createType({
+          ... prop,
+          sourceProperty: prop
+        }, true)
```
bterlson, line 446: we don't need to put these property types in the cache because there's no way to reference them by node anyway. A node never existed.

packages/adl/compiler/parser.ts
```diff
@@ -151,31 +151,40 @@ export function parse(code: string | Types.SourceFile) {
       parseExpected(Token.GreaterThan);
     }
 
-    if (token() === Token.OpenBrace) {
-      parseExpected(Token.OpenBrace);
-      const properties = parseModelPropertyList();
-      parseExpected(Token.CloseBrace);
+    if (token() !== Token.Equals && token() !== Token.OpenBrace && token() !== Token.ExtendsKeyword) {
+      throw error('Expected equals or open curly after model statement');
```
bterlson, line 155: Need to update this to include `extends`

packages/adl/compiler/parser.ts
```diff
@@ -600,19 +618,19 @@ export function parse(code: string | Types.SourceFile) {
...
-  ): A | B {
-    if (token() == expectedTokenA) {
-      nextToken();
-      return expectedTokenA;
-    } else if (token() == expectedTokenB) {
-      nextToken();
-      return expectedTokenB;
-    } else {
-      throw error(`expected ${Token[expectedTokenA]} or ${Token[expectedTokenA]}, got ${Token[token()]}`);
+  function parseExpectedOneOf<T extends Token[]>(... options: T): T[number] {
```
bterlson, line 621: Turned out I didn't need this but I like this generalization of parseExpectedOneOf so maybe it can stay? 😁

packages/adl/samples/visibility/visibility.adl
```diff
@@ -0,0 +1,31 @@
...
+model Readable<T> {
+  ... T;
+}
+
+@wv("write")
+model Writable<T> {
+  ... T;
+}
+
+model ReadablePerson {
```
bterlson, line 19: Would love to not type these, see #292

packages/adl/test/output/visibility/openapi.json
```diff
@@ -0,0 +1,114 @@
+{
```
bterlson, line 1: This is pretty darn nice I think 😁

packages/adl/test/test-scanner.ts
```diff
@@ -1,6 +1,7 @@
 import { deepStrictEqual, strictEqual } from 'assert';
 import { readFile } from 'fs/promises';
 import { URL } from 'url';
+import { isMainThread } from 'worker_threads';
```
bterlson, line 4: auto import hosing again

packages/adl/compiler/cli.ts
```diff
@@ -71,14 +71,14 @@ const args = yargs(process.argv.slice(2))
   .argv;
 
 async function compileInput(compilerOptions: CompilerOptions): Promise<void> {
-  try {
+  //try {
```
bterlson, line 74: I'll undo this (but would probably prefer that --debug not even wrap in try/catch so break on uncaught exceptions works - separate issue 😀)

packages/adl/compiler/checker.ts
```diff
@@ -394,6 +416,55 @@ export function createChecker(program: Program) {
     }
   }
 
+  function checkClassHeritage(heritage: ReferenceExpression[]): ModelType[] {
+    return heritage.map(heritageRef => {
+      const heritageType = getTypeForNode(heritageRef);
+
+      if (heritageType.kind !== "Model") {
+        throw new Error("Models must extend other models.")
```
daviwil, line 424: Should we also check to make sure the model type isn't intrinsic?  What happens if someone `extends string`?

packages/adl/lib/openapi.ts
```diff
@@ -606,15 +623,15 @@ function createOAPIEmitter(program: Program, options: OpenAPIEmitterOptions) {
   }
 
   function getSchemaForModel(model: ModelType) {
-    model = getTypeForSchemaProperties(model);
+    model = getTypeForSchemaProperties(model);``
```
daviwil, line 626:
```suggestion
    model = getTypeForSchemaProperties(model);
```

packages/adl/samples/visibility/vis.js
```diff
@@ -0,0 +1,24 @@
+const visibility = new Map();
+export function getVisibility(entity) {
+  return visibility.get(entity);
+}
+export function v(c, t, v) {
```
daviwil, line 5: I've got a @visibility decorator defined in lib/decorators.ts:

https://github.com/daviwil/adl/blob/master/packages/adl/lib/decorators.ts#L140

Should the rest of this code go there too?

packages/adl/test/output/testserver/media-types/openapi.json
```diff
@@ -54,7 +57,9 @@
         "responses": {
           "200": {
             "schema": {
-              "type": "string"
+              "type": "object",
```
daviwil, line 60: How'd this change?

daviwil (approved):
Do you think that `OrganizationResource` will be fixed once we start using `extends` on the inheritance from `ArmTrackedResource<T>`?

packages/adl/compiler/checker.ts
bterlson: I think it's ok. You don't inherit the "intrinsic" part (since the derived class won't have the metadata). I think the real question is, what happens if you try to add properties to a primitive. We don't really have an opinion on that yet!

packages/adl/samples/visibility/vis.js
bterlson: Ah sure I can move it into there. I'll go with `visibility`/`withVisibility` too.

packages/adl/test/output/testserver/media-types/openapi.json
bterlson: Looking forward to finding out!

packages/adl/samples/visibility/vis.js
daviwil: Cool!  That would complete issue #233.

packages/adl/test/output/visibility/openapi.json
daviwil: Yeah, it looks great!

packages/adl/test/output/testserver/media-types/openapi.json
daviwil: I think it must be because `OkResponse<T>` is spreading whatever type is placed in the template param.  If we switch that to `extends T` then it should be fine.  Now I see that we have to allow extending intrinsics in that case :)

packages/adl/test/output/petstore/openapi.json
```diff
@@ -30,10 +30,10 @@
             }
           },
           "default": {
+            "description": "Error",
```
bterlson, line 33: The changes in this file are just reordering. Description is first now is all.

packages/adl/compiler/parser.ts
nguerrera: Neat, why not!

packages/adl/samples/visibility/vis.js
nguerrera: (Redundant +1 for `visibility / withVisibility.`. I was very perplexed looking at above. Give the parameters good names too, please. 😁 )

nguerrera (approved):

packages/adl/test/output/confluent/openapi.json
```diff
@@ -402,12 +287,11 @@
                   "type": "string"
                 }
               },
-              "description": "Paged response",
               "required": [
                 "value",
                 "nextLink"
               ],
-              "x-adl-name": "Page<OrganizationResource>"
+              "x-adl-name": "<OrganizationResource>"
```
daviwil, line 294: Interesting that this changed, though it doesn't really matter much

bterlson: Kinda mad about it... but are you saying it's ok for now?

PR #293, opened 2021-02-11
Add body time sample

timotheeguerin:
For this one we are missing the `time` type.

PR #294, opened 2021-02-11
Add `time` intrinsic type

daviwil:
This PR adds a new intrinsic type called `time` for representing a time without a date.  AutoRest accepts this as a `type: string, format: time` schema which ultimately gets used as a time-formatted string in code generators.

Adding this to unblock @timotheeguerin's PR #293.


timotheeguerin (approved):

nguerrera:
Is it expected that this also includes namespace changes?

packages/adl/compiler/checker.ts
```diff
@@ -5,10 +5,10 @@ import {
   BooleanLiteralNode,
   BooleanLiteralType,
   IdentifierNode,
-  NamespacePropertyNode,
+  OperationStatementNode,
```
nguerrera, line 8: I think maybe you meant to leave this to another branch/pr, but let's also update the grammar to match these changes. Should be able to take what I have in the proposal and #274.

nguerrera (approved):
The time change looks good to me. The namespace change looks fine too apart from wanting the .grammar change to come with it.

packages/adl/compiler/parser.ts
```diff
@@ -84,16 +84,36 @@ export function parse(code: string | Types.SourceFile) {
...
+          continue;
+        case Token.NamespaceKeyword:
+          statements.push(parseNamespaceStatement(decorators));
+          continue;
+        case Token.OpKeyword:
+          statements.push(parseOperationStatement(decorators));
+          continue;
+      }
+
+      throw error(`Expected statement, but found ${Token[tok]}`);
```
nguerrera, line 116: Seems like we could factor out a common parseStatementList for the top-level and inside namespace.

packages/adl/compiler/parser.ts
```diff
@@ -84,16 +84,36 @@ export function parse(code: string | Types.SourceFile) {
       }, modelPos);
     }
 
-
```
nguerrera, line 87: Not here, but on line 75 above looks like we had some syntax like `interface X (param: Type, ...) {` and that's carrying over to namespace. I forgot we had that but I don't think it should stay. I didn't document it in grammar and I don't think we use it anywhere, but I could be mistaken.

nguerrera (changes requested):
Found a few other small things in namespaces and it's probably better to split into two PRs.

bterlson:
Would `plainTime` be a better name than `time`? `time` feels a bit ambiguous to me, and confusing `time` with `timestamp`, zoned date time, or local date time.

daviwil:
Holy crap, I thought I had stashed the namespace changes, sorry about that!  I'll pull them out tomorrow.

nguerrera:
No worries. Been there!

packages/adl/compiler/parser.ts
daviwil: I'll take it out!  Doesn't make as much sense for namespaces, for sure.

nguerrera (approved):

PR #295, opened 2021-02-12
Log source location with all errors

### LINKED ISSUE #275: Initial pass at better error messages
We haven't made any effort yet to have a decent experience with diagnostics. This would be a first step towards that.

* Keep track of source files
* Make sure the thrown errors are surfaced with corresponding source file and line.

NB: This issue does not track error recovery yet, we will still bail at first error, just with a more and better info.

nguerrera:
Fix #275

* Everywhere we throw an error now reports a location. Use `throwDiagnostic` plus getSourceLocationOfXxx helpers going forward for new errors. (We still need to do error recovery and not just throw at the first sign of trouble, but that isn't part of this initial pass.)

* Fix duplicate name reporting in binder to report duplicates in the same file as an error and to report locations of all duplicates up to the first bound source file that introduced at least one. This found an issue in one of the samples! 😁

* Fix issue where failed compilation was saying "Compilation completed successfully".


packages/adl/compiler/cli.ts
```diff
@@ -107,11 +115,14 @@ async function main() {
...
-    await compileInput(options);
-    console.log(`Compilation completed successfully, output files are in ${options.outputPath}.`)
+    if (await compileInput(options)) {
+      console.log(`Compilation completed successfully, output files are in ${options.outputPath}.`)
+    }
   } else if (args._[0] === "generate") {
     const options = await getCompilerOptions();
-    await compileInput(options);
+    if (!(await compileInput(options))) {
+      return;
```
nguerrera, line 124: Note to self, check the process exit code.

daviwil (approved):
Looks fantastic!

nguerrera:
Thanks, @daviwil !

Can you look at https://github.com/Azure/adl/pull/295/commits/be270d18cd71e66ee4c6f4ddc42818ff36b70201 too. I just pushed it.

packages/adl/compiler/cli.ts
```diff
@@ -146,8 +157,12 @@ async function main() {
...
-    console.error(`An unknown error occurred:\n\n${err.message}`);
-    if (args["debug"]) {
-      console.error(`Stack trace:\n\n${err.stack}`);
-    }
+    // NOTE: An expected error, like one thrown for bad input, shouldn't reach
+    // here, but be handled somewhere else. If we reach here, it should be
+    // considered a bug and therefore we should not suppress the stack trace as
+    // that risks losing it in the case of a bug that does not repro easily.
+    console.error("Internal compiler error!");
+    console.error("File issue at https://github.com/azure/adl\n");
```
bterlson, line 165: nice.

bterlson (approved):
Think it would be nice to make throwDiagnostic easier to call by taking the reference node/type rather than source pointer, but can be done later!

packages/adl/compiler/cli.ts
nguerrera: I'm so relieved you like this. I have had many arguments on this over the years, and it is one of very few hills I'm prepared to die on. 😁

nguerrera:
@bterlson Thanks, that's a great idea and I'll put together a PR with it first thing next week.

PR #298, opened 2021-02-12
Add support for nested namespaces

daviwil:
This change updates the ADL parser to allow nested namespace statements in the language syntax.  I've also updated the language grammar to reflect the changes up until now.  I've also removed the parsing of the old interface parameter syntax which we aren't supporting anymore.

One thing I don't know is whether I need to wire up the nesting in some way to create new scopes.  I'm still looking into this, but if someone knows off the top of their head, it'd be helpful!


packages/adl/test/output/nullable/openapi.json
```diff
@@ -47,6 +47,7 @@
         ],
         "responses": {
           "200": {
+            "description": "",
```
daviwil, line 50: I changed the response generation code a few days ago to discard empty descriptions but it turns out AutoRest throws an error if you don't include one!  I've fixed this now and regenerated to add them back.

packages/adl/compiler/types.ts
```diff
@@ -13,8 +13,8 @@ export type Type =
   | ModelType
   | ModelTypeProperty
   | TemplateParameterType
-  | Namespace
```
daviwil, line 16: Named this incorrectly in the last namespace PR!  Fixed it now.

PR #300, opened 2021-02-13
Fix class extends T

bterlson:
I found that the following didn't work:

```
class Foo<T> extends T { }
```

At first because I didn't implement the visitor for the heritage clause, and then because class evaluation didn't take into account the possibility of uninstantiated template parameters. Since we're not using uninstantiated classes for anything beyond light semantic checks if anything (AFAIK), this PR just elides them from the heritage list.

If in the future we need to do some calculations on uninstantiated template parameters in heritage (e.g. in the presence of generic constraints) we'll have to do something different. For now, this works IMO.


packages/adl/compiler/diagnostics.ts
```diff
@@ -36,7 +37,8 @@ export function getSourceFileOfNode(node: Node): SourceFile {
     node = node.parent;
   }
   if (node.kind !== SyntaxKind.ADLScript) {
-    throw new Error("Cannot obtain source file of unbound node.");
```
daviwil, line 39: Needs to be reverted?

packages/adl/compiler/cli.ts
```diff
@@ -29,6 +29,11 @@ const args = yargs(process.argv.slice(2))
           type: "string",
           default: "./adl-output",
           describe: "The output path for generated artifacts.  If it does not exist, it will be created."
+        })
+        .option("skip-lib", {
```
daviwil, line 33: Is there a reason why a user might want to do this?  Wondering if there's a more robust way we can achieve it longer-term

daviwil (approved):

packages/adl/compiler/cli.ts
bterlson: Oh whoops I didn't mean to include this, but put it in a separate PR. I'll pull it out of here.

The main reason is that say I want to put a breakpoint on template instantiation - since the std lib comes with many such templates, I have to continue and inspect to see if I'm on the instantiation I care about. Easier to repro simple cases with skip-lib.

nguerrera: I think this is good to have. My only comment is on the naming. I like "nostdlib", which C# and gcc both use IIRC.

nguerrera (approved):

nguerrera:
I only mention this because it bugged you the last time you forgot: you probably will want to squash merge this after the fixups.

packages/adl/compiler/cli.ts
daviwil: `nostdlib` sounds good!

daviwil:
Should we just make squash and merge the default?

nguerrera:
I don't know how to do it, but go for it if it's possible. It picks the last thing you used per repo otherwise as the default IIRC.

nguerrera:
We can require squash merge, but that feels bad to me.

daviwil:
Ahhh, well if it remembers the last thing you chose, maybe that's good enough?

bterlson:
I will ensure to pick the squash and merge button this time, and hopefully my problems vanish 😭

